PR #12, opened 2024-05-16
Finish updating FlexMeasures host url

Flix6x:
You had already updated the host url, but to avoid any confusion, I also updated the url in commented-out lines.


ArdJonker (approved):
Good idea!

Issue #17, opened 2024-05-31
Missing instructions for timezone key in secrets file


ArdJonker:
AppDaemon is dependant on timezone information. We store this in the secrets file. The instructions for this are missing in the docs.md.
Also there is an incorrect version number in UI.

PR #18, opened 2024-05-31
#17 Missing instructions for editing secrets file.

ArdJonker:


ArdJonker:
Small change so skipped review.

PR #44, opened 2024-07-16
43 improve detection of communication failure with charger

ArdJonker:


ArdJonker (commented):
Self reviewed, all is oke.

PR #45, opened 2024-08-06
Simplify folder structure

rhpijnacker:
We are creating a dummy appdaemon folder in the target location, only to immediately overwrite it with the correct content.
Because of this, there is a `appdaemon` and a `v2g-liberty/appdaemon` folder in the addon repository.

We can simplify this by moving both the `v2g-liberty/appdaemon` and `v2g-liberty/homeassistant` folders up by one level.
(This means we no longer need the `v2g-liberty` sub-folder.)


ArdJonker (approved):
Is logical and a nice clean-up. Has run in test environment for half a day without problems.
Seems based on version 0.1.11 though. Will fix at merge.

Issue #48, opened 2024-08-07
Modbus possibly more stable when not closing connection after every call


ArdJonker:
There are still many crashes of the modbus module in the Quasar.
@rhpijnacker has build a modbus client that seems to result in more stable communication and this does not close the connection after every call and does not prevent parallel calls.

PR #49, opened 2024-08-07
#48 Modbus possibly more stable when not closing connection after every call.

ArdJonker:
- Just opening the connection, not closing anymore. Only when configuration changes.
- Removed function to always have one call open at a time as well.
- Reverting to defaults for modbus client timeout, retries, reconnect_delay and _max.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
@@ -169,7 +169,7 @@ class ModbusEVSEclient(hass.Hass):
     CHARGER_SET_CHARGE_POWER_REGISTER: int = 260
 
     # AC Max Charging Power (by phase) (hardware) setting in charger (Read/Write)
-    # (int16) unit W, min 1380, max 7400
+    # (int16) unit W, min_value 1380, max_value 7400
     # Used when set_setpoint_type = power
     MAX_AVAILABLE_POWER_REGISTER: int = 514
     # The Quasar does not accept a setting lower than 6A => 6A*230V = 1380W
@@ -207,10 +207,9 @@ class ModbusEVSEclient(hass.Hass):
     AVAILABILITY_STATES = [1, 2, 4, 5, 11]
 
     # Modbus variables
-    client: modbusClient
+    client: modbusClient = None
     WAIT_AFTER_MODBUS_WRITE_IN_MS: int = 2500
     WAIT_AFTER_MODBUS_READ_IN_MS: int = 50
-    modbus_connection_in_use: bool = False
 
     # For sending notifications to the user.
     v2g_main_app: object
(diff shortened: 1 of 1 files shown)


ArdJonker:
It has run on test and production for half a day. I'll wait an other two days before merging.

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
```diff
@@ -274,20 +273,22 @@ async def initialize(self):
 
     async def initialise_charger(self, v2g_args=None):
         self.log(f"Configuring Modbus EVSE client at {c.CHARGER_HOST_URL}:{c.CHARGER_PORT}, reason: {v2g_args}")
+
+        # Remove old client if needed.
+        if self.client is not None:
+            if self.client.connected:
+                self.client.close()
+            self.client = None
```
rhpijnacker, line 281: Setting the client to `None` has little added value. In the next line you are assigning it a new value. I think this line can be safely removed.

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
```diff
@@ -1115,11 +1034,8 @@ async def __force_get_register(self, register: int, min: int, max: int) -> int:
 
             if result is not None:
                 try:
-                    result = int(float(result.registers[0]))
-                    if result > self.HALF_MAX_USI:
-                        # This represents a negative value.
-                        result = result - self.MAX_USI
-                    if min <= result <= max:
+                    result = self.__get_2comp(result.registers[0])
```
rhpijnacker, line 1037: This looks like an unrelated change. Did you mean to include this?

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
```diff
@@ -274,20 +273,22 @@ async def initialize(self):
...
         self.client = modbusClient.AsyncModbusTcpClient(
             host=c.CHARGER_HOST_URL,
             port=c.CHARGER_PORT,
-            timeout=3,
-            retries=3,
-            reconnect_delay=500,
-            reconnect_delay_max=1000,
             retry_on_empty=True,
         )
         await self.client.connect()
```
rhpijnacker, line 288: I would expect some error handling for the scenario when the `connect()` call fails. Or is that handled somewhere else and did I miss it?

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
ArdJonker: Will do

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
ArdJonker: You are right, unrelated but relevant. Will take more care is splitting up in the future.

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/modbus_evse_client.py
ArdJonker: Strangely enough the (connection-) errors occur not while connecting but while reading / writing. So that is where the error handling is now is focussed.

PR #73, opened 2024-09-12
Make python dependencies in Dockerfile versionless

rhpijnacker:
Recently we had some problems with building the Docker container. 
This is caused by an updated of the python packages in the alipine repository that this addon builds on top of.
Since the Dockerfile mentions explicit versions, these versions are no longer available after the update of the external repository, causing this build problem.
Since the only way to fix the problem, is to update the Dockerfile with the new version, I do not see the added value of explicitly naming the version.


ArdJonker:
I'll try it out on local dev and test.

PR #85, opened 2024-09-19
84 bug getting epex prices on daily schedule fails

ArdJonker:


ArdJonker:
I am not a 100% sure if this has solved the problem. Testing is somewhat hard and i did run the kick_off_daily methods many times in test setting i did not let it run form several days to see if all is well.

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/v2g_liberty.py
```diff
@@ -862,23 +860,29 @@ async def __set_soc_prognosis_in_chart(self, chart_line_name: ChartLine, records
...
         # To make sure the new attributes are treated as new we set a new state as well
-        new_state = "SoC prognosis at " + now.isoformat()
+        new_state = "Chart line data at " + now.isoformat()
+        entity = f"input_text.{self.chart_line_entity[chart_line_name]}"
 
         if records is None:
-            await self.set_state(f"input_text.{self.chart_line_entity[chart_line_name]}",
-                                 state=new_state, attributes=clear_line_records)
-            self.log(f"__set_soc_prognosis_in_chart chart_line_name: '{chart_line_name}' cleared.")
+            await self.set_state(entity, state=new_state, attributes=clear_line_records)
```
rhpijnacker, line 867: If I look at the documentation `set_state`, then this appears not to be an `async` method. Are you sure you want to use `await` here?

rhpijnacker (approved):

v2g-liberty/rootfs/root/appdaemon/apps/v2g-liberty/v2g_liberty.py
ArdJonker: In the documentation it does not have any info on async behaviour of methods.
It is however necessary to await set_state in async methods (learned the hard way...).

Issue #84, opened 2024-09-18
🪲 BUG: Getting EPEX prices on daily schedule fails


ArdJonker:
Maybe the the checks for "now renewed" and await statements for run_daily may be the cause, but not sure...

Issue #71, opened 2024-08-29
"Reduce max. (dis-)charge power" setting not preserved


rhpijnacker:
I noticed that with the Load Balancer enabled, the charger limits the maximum power to around 4000W even though it is set to 20A (4750W).
So help FM making better schedules, I decided to enter these values in the "Reduce max. (dis-)charge power" fields. However, it seems that after restarting the add-on, this setting always defaults back to disabled.

ArdJonker:
The base problem is that the toggle "Reduce max discharge power" is not preserved. And that is a wider problem: none of the toggles are preserved. So also VAT & Markup and change default server.

PR #86, opened 2024-09-19
#71 🪲 BUG: 'Reduce max. (dis-)charge power' setting not preserved

ArdJonker:
This was due to incomplete list of 'True values' in the __check_and_convert_value method. Also fixed:
- Different FlexMeasures URL can now be used and remains valid over restarts.
- VAT and Markup can be used. For all optional values: the content of the input-fields is saved if toggle is switched off.

***Please note***
Somehow a lot of changes from bug #84 'Getting EPEX prices on daily schedule fails' had to be re-commited (wrong base branch?)


ArdJonker (commented):
Self approved

rhpijnacker (approved):

PR #95, opened 2024-09-26
PATCH for minor bugs in octopus_price_data_manager and modbus_evse_client

ArdJonker:
As this is relevant to specifically one user i decided not to make a new version in config.yaml

- Fixed wrong name self.second_try_time_get_data for octopus_price_data_manager.py
- Removed blocking set_charge_power in modbus_evse_client.py as this blocks try_get_soc.


ArdJonker (commented):
Self approved

Issue #100, opened 2024-09-26
🪲 BUG: Error message for "Charger is controlled by app" not showing


ArdJonker:
This is due to wrong name of charger status 3 in modbus_evse_client.py

PR #101, opened 2024-09-26
#100 🪲 BUG: Error message for "Charger is controlled by app" not showing

ArdJonker:
- Changed the name of the state 3 to "Connected: controlled by Wallbox App"
- Did some cleanup in v2g_liberty module
- did some cleanup in dashboard


ArdJonker (commented):
Self approved

Issue #103, opened 2024-09-27
TECH: Use get_plugin_config() instead of template sensor TZ in package.yaml


ArdJonker:
get_plugin_config() provides:
- Currency
- Timezone
- Length (km)
- Name of HA instance (replace from v2g_liberty.py)
- Country (pre-select for electricity-providers?)

Use in globals:
```python
config = await self.get_plugin_config()
```

Replaces:
```yaml
template:
  - sensor:
    # To make the HA time_zone available to the V2G Liberty python code.
    - name: Time Zone
      unique_id: time_zone
      state: "{{ now().tzinfo }}"
```

PR #107, opened 2024-10-02
 #103 TECH: Use get_plugin_config() instead of template sensor TZ in …

ArdJonker:
…package.yaml

 - For HA_NAME and time_zone this in fixed.


ArdJonker (commented):
Self approved

Issue #108, opened 2024-10-02
REFACTOR: Restructure code in set_next_action


ArdJonker:
It is hard to read and needs tweaks for structure.

PR #109, opened 2024-10-02
 #108 REFACTOR: Restructure code in set_next_action

ArdJonker:
- This makes recovering from invalid SoC more robust
 - BUMP own version to 0.3.2


ArdJonker (commented):
Self approved

Issue #113, opened 2024-10-05
TECH: Add "initialised" attribute to settings entities (=preparation for UI improvements)


ArdJonker:
So that development of custom cards can be testing with this.

PR #114, opened 2024-10-05
 #113  TECH: Add "initialised" attribute to settings entities (=preparation for UI improvements)

ArdJonker:
Also changed upper limit for max-charge from 100(%) to 95 as the Leaf will never charge above 97% via CHAdeMo.


ArdJonker (commented):
Self approved

Issue #124, opened 2024-10-17
🪲 BUG: admin_mobile name select is changing several times per second


ArdJonker:
The new release somehow triggers a loop in the code for setting the options in a input_select.
Restarting HA sometimes helps but it every now and then is needed to do several restarts.

Because we expect a new frontend for the setting in one of the next releases we temporary will always remove the `Please choose an option` option.

PR #125, opened 2024-10-17
#124 🪲 BUG: admin_mobile name select is changing several times per second

ArdJonker:
- Removed "Please choose an option" from all input-select elements.
- Removed the persistent notification for missing admin mobile name.
- Further split the changelog.md to hold only changes for the current release and added changelog_for_all_releases.md for a complete overview of all changes.


ArdJonker (commented):
Self approved

Issue #128, opened 2024-11-01
Feature Request: Make chart two columns wide on desktop


ArdJonker:
Also replace SoC gauge with progress-bar for more efficient use of space

PR #129, opened 2024-11-01
 #128 Feature Request: Make chart two columns wide on desktop.

ArdJonker:
- Split dashboard into two files for readability (settings view has not changed further)
- Changed the main view to a sections view
- Made the chart two columns wide
- Changed the SoC Gauge into a progressbar to save space
- Moved the stats to bottom right of the view
- Minor text changes in view
- Fixed the hover on Pause button

- Added the code of #124 as this was not on dev yet (only on the main branche), sorry for bad git work..).


ArdJonker (commented):
Self approved

Issue #130, opened 2024-11-01
🪲 BUG: Alert in UI for "Price and/or emission data imcomplete" never shown


ArdJonker:
The get_fm_data only caters for the notification and does not set the input_boolean.error_epex_prices_cannot_be_retrieved that is intended for the alert in the UI.

PR #131, opened 2024-11-01
#130 🪲 BUG: Alert in UI for "Price and/or emission data incomplete" n…

ArdJonker:
…ever shown

- Added the method `set_price_is_up_to_date` on the v2g_liberty.py module to set `input_boolean.error_epex_prices_cannot_be_retrieved` so that get_fm_data.py module does not have to interact with the UI.
- From the get_fm_data this method is called.

- Made code for `notify_user_of_charger_needs_restart()` more logical (not really part of this issue).

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.4.0 2024-10-30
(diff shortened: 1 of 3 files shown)


ArdJonker (commented):
Self approved

Issue #136, opened 2024-11-02
 🛠️ REFACTOR: rename folder v2g-liberty to v2g_liberty for better compatibility with pytest


ArdJonker:
In preparation of using pytest

PR #137, opened 2024-11-02
#136 🛠️ REFACTOR: rename folder v2g-liberty to v2g_liberty for better…

ArdJonker:
… compatibility with pytest


ArdJonker (commented):
Self approved

PR #144, opened 2024-11-05
Add local calendar to init-config

rhpijnacker:
As requested, added the initial files in `config/.storage` with a local calendar called "Car reservation".


ArdJonker (approved):
Looks fine to me!

Issue #148, opened 2024-11-07
🪲 BUG: Chartlines for prices / emissions do not extrand to the periode they represent.


ArdJonker:
The value of 23:00 should be extended in the chart to 00:00:

![afbeelding](https://github.com/user-attachments/assets/9061f8a1-d50b-4968-8016-c27f849606d1)

PR #149, opened 2024-11-07
 #148 🪲 BUG: Chart lines for prices / emissions do not extend to the period they represent.

ArdJonker:
Added only for the prices as emissions have a 5 min. resolution and are already added to 23:55.


ArdJonker (commented):
Self approved

PR #150, opened 2024-11-07
Hot-fix for failing chart update

rhpijnacker:
This fixes the problems with updating the chart.


v2g-liberty/rootfs/etc/s6-overlay/s6-rc.d/init-appdaemon/run
ArdJonker, line 1: This indeed needs to be deleted!

ArdJonker (approved):

Issue #152, opened 2024-11-08
🪲 BUG: In Max charge the power is set to 0 every now and then


ArdJonker:
This is likely caused by old schedule still running.

PR #153, opened 2024-11-08
#152 🪲 BUG: During Max charge the power is set to 0 every now and then

ArdJonker:
- When the charge-mode changes from Automatic to Max charge the charge timers are cancelled.
- Added 'source' to some method calls for debugging.
- Made the Automations for setting the charge-mode in the UI comply with the new (HA 2024.10) style yaml. This makes this new version of HA required.

- Also added a more practical max SoC for the car battery of 97% as the car will never charge higher than that (partly related).
- Also prevented a exception for a None result for getting charged energy (not related).


ArdJonker (commented):
Self approved

Issue #147, opened 2024-11-07
REFACTOR: Use sections precision mode for card sizes in UI for a.o. `Automatic`, `Charge now` and `Pause` buttons


ArdJonker:
See:
https://www.home-assistant.io/blog/2024/11/06/release-202411/#precise-mode-for-card-sizes

PR #154, opened 2024-11-08
#147 REFACTOR: Use sections precision mode for card sizes in UI for a…

ArdJonker:
….o. Automatic, Charge now and Pause buttons

- Also removed "initial: true" on chargemodeautomatic as this was undesired behaviour.
- Also renamed Disconnect to Unplug for more clarity and smaller text to fit narrow button.


ArdJonker (commented):
Self approved.
Some stuff from older PR's? Don't know where these came from..?

PR #160, opened 2024-11-13
Format using VS Code formatter

rhpijnacker:
I simply opened every file and saved it using VS Code's formatter


ArdJonker (approved):
Looks very structured. Did not locally test.

PR #165, opened 2024-11-14
Feature refactor python classes

rhpijnacker:
To make the V2G Liberty classes testable, it helps if they do not subclass from the AppDaemon classes, and that manage the dependencies between the classes by explicitly injecting them.
This makes it a lot easier to mock the class dependencies and start testing.


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/log_wrapper.py
ArdJonker, line 1: 👍🏻

ArdJonker (approved):
All fine.
Tested the`res.get_loop().call_soon_threadsafe(res.set_result, data["response"])` on the test environment and it works nicely.

PR #166, opened 2024-11-15
As the appdaemon logger now only adds v2g_app and no further details …

ArdJonker:
…of where the log_statement came from i added the Module, line_number and Function in a private `__log` method in ReservationsClient.

ToDo:
Make this a generic method/function in Globals for all modules to use.

I also added `missing_app_warnings: 0` to the `appdaemon.yaml` to get rid of (unneeded) warnings of all the classes that are no longer separate app-daemon apps.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/appdaemon.yaml
+++ b/v2g-liberty/rootfs/root/appdaemon/appdaemon.yaml
@@ -10,6 +10,9 @@ appdaemon:
   # If you would like your log timestamps to be in local time set it to the right timezone here.
   time_zone: Europe/Amsterdam
 
+  # Suppress warnings in the log for missing app configuration (in the apps.yaml).
+  missing_app_warnings: 0
+
   production_mode: True
   plugins:
     HASS:
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
@@ -2,7 +2,7 @@ import datetime as dt
 import re
 import requests
 import constants as c
-import inspect
+import log_wrapper
 from v2g_globals import he, get_local_now
 import caldav
 from service_response_app import ServiceResponseApp
@@ -30,6 +30,7 @@ class ReservationsClient:
 
     def __init__(self, hass: Hass):
         self.hass = hass
+        self.__log = log_wrapper.get_class_method_logger(hass.log)
 
     async def initialize(self):
         self.__log("initialise ReservationsClient")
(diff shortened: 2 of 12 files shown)


v2g-liberty/rootfs/root/appdaemon/appdaemon.yaml
```diff
@@ -10,6 +10,9 @@ appdaemon:
   # If you would like your log timestamps to be in local time set it to the right timezone here.
   time_zone: Europe/Amsterdam
 
+  # Suppress warnings in the log for missing app configuration (in the apps.yaml).
+  missing_app_warnings: 0
```
rhpijnacker, line 14: Nice! I was planning on adding this, but good to see you already did it.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
```diff
@@ -225,14 +226,28 @@ async def activate_selected_calendar(self):
...
         #          f"every {self.POLLING_INTERVAL_SECONDS} sec.")
-        self.hass.log("Completed activate_selected_calendar")
+        self.__log("Completed activate_selected_calendar")
 
     ######################################################################
     #                   PRIVATE (CALLBACK) FUNCTIONS                     #
     ######################################################################
 
+    def __log(self, msg:str, level:str=""):
+        info = inspect.stack()[1][0]
```
rhpijnacker, line 238: Do you know how "expensive" it is to do this?

For the purpose of adding the class-name, I think we can fix that by initialising a log function in the constructor with a "partial" that captures the class-name.
But, ofcourse, then you will not have the benefit of also automatically capturing the method name.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_globals.py
```diff
@@ -1686,7 +1686,6 @@ async def __cancel_setting_listener(self, setting_object: dict):
 #                           UTIL FUNCTIONS                           #
 ######################################################################
 
-
```
rhpijnacker, line 1689: Was this an automated formatting change?
Maybe we should set up a pre-commit hook to make sure we do not keep on changing white space for no good reason?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
rhpijnacker: https://gist.github.com/JettJones/c236494013f22723c1822126df944b12 may give a good indication on what the "best" way to do this is.

rhpijnacker:
> Make this a generic method/function in Globals for all modules to use.

Would it make sense to start splitting "globals" into smaller modules? This module seems a bit too stuffed with all kinds of utility functions.
E.g. I would look for a function like the one you added in a module called "logging.py".

rhpijnacker:
I updat

Issue #167, opened 2024-11-18
Local Home Assistant calender only triggers changed event on first opcoming calendar item


ArdJonker:
So maybe the polling that is done for remote calendars should also be done for local calendar besides working with the changed event.

ArdJonker:
Resolved by branch #158-1

Issue #158, opened 2024-11-12
Feature Request: Notify user when at the end of a calendar item the car is not connected.


ArdJonker:
Give it a slack of an hour or so (Later: make this a user-setting)

ArdJonker:
Implemented by branch #158

ArdJonker:
The changes seem to have been lost in the github processes (i probably did this... 😣)

ArdJonker:
Resolved by PR #168

PR #168, opened 2024-11-19
#158-2 Feature request: Notify user when at the end of a calendar item…

ArdJonker:
This is a re-apply of the code that was supposed to be in the #158 branch. Something has gone wrong with the PR.

It also includes:
- Fix for issue #167: Local Home Assistant calendar only triggers changed event on first upcoming calendar item.
  The local calendar is now also polled.
- An update for the log_wrapper to produce better log lines.


ArdJonker (commented):
Self approved

Issue #169, opened 2024-11-20
🪲 BUG: Alert for negative prices not working correctly


ArdJonker:
In the get_fm_data.py module the method "__check_negative_price_notification" is not called when a negative price is detected.

PR #171, opened 2024-11-21
 #169 🪲 BUG: Alert for negative prices not working correctly

ArdJonker:
- Fixed problem
 - Also fixed situation where consumption and production prices are different.
 - Cleanup of code


ArdJonker (commented):
Self approved

PR #172, opened 2024-11-23
Feature-ping-pong

rhpijnacker:
This PR adds a "ping-pong" interface to the V2G-Liberty addon, with an accompanying `custom:v2g-liberty-ping-card` for the frontend.

This card is invisible by default.
It will fire a `ping` event every `interval` seconds and expects a `ping.result` event back.
If that does not happen within the `ping_timeout`, then it will show an alert informing the user the V2G-Liberty is not responding.

This PR also introduces a new dev-container for the V2G-Liberty frontend cards. 
To make this happen, I had to do some structural changes, moving the .devcontainer folder one level up. 
This will require you to recreate the containers.


v2g-liberty/rootfs/root/homeassistant/www/v2g_liberty/v2g-liberty-cards/v2g-liberty-cards.js
rhpijnacker, line 1: The devcontainers are set-up such that when you make changes to one of the cards in the "v2g-liberty-cards" container, it will automatically publish them to Home Assistant so you will only have to reload (twice) to get the changes.
It will also update the compiled `v2g-liberty-cards.js` and accompanying `.map` (used for debugging) in the folder that the appdaemon addon will copy to end-user's deployments.

v2g-liberty-cards/src/strings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "ping-card": {
+        "error": "V2G-Liberty is not responding",
```
rhpijnacker, line 3: Fee free to suggest other UI texts.

v2g-liberty-cards/src/strings.json
ArdJonker, line 1: To bring it in line with other error messages i suggest the following. As we cannot really explain what the problem is i just refer to it as an error. 
The main message should be what can the user do to fix this problem, that is what i added.
I hop this sort of looks like the this:
afbeelding


Title:
V2G Liberty error 

Message:
There is an error in V2G Liberty.

**You can fix this:**

1. Restart V2G Liberty add-on<br/>
   `Main menu` > `Settings` > `Add-ons` > `V2G Liberty` > `Restart`

2. Restart Home Assistant<br/>
    `Main menu` > `Settings` > `⋮` (top right) > `Restart Home Assistant` (3x)

v2g-liberty-cards/src/util/appdaemon.ts
ArdJonker, line 1: Ping every minute seems a bit frequent? Probably 'costs' are low, then it's oke.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
ArdJonker, line 1: 🏓😄

ArdJonker (approved):
Lot's of files i view i cannot really judge ...
Added a suggestion for the text for the error message.

v2g-liberty-cards/src/strings.json
rhpijnacker: I do not think it add much value to restart Home Assistant, and I would definitely recommend restarting V2G-Liberty _after_ restarting Home Assistant as a general rule.

rhpijnacker:
> Lot's of files i view i cannot really judge ... Added a suggestion for the text for the error message.

I suggest you take a look at the separate commits I did. I tried to make an effort to make them as understandable as possible.

v2g-liberty-cards/src/util/appdaemon.ts
rhpijnacker: The 60 seconds is the default timeout value for "function calls" to the app-daemon.
I actually override this for the ping-timeout to a default of 5 seconds (which should be more than enough).

The interface is very low-cost indeed, so I propose to do this with an interval of 15 sec.

v2g-liberty-cards/src/strings.json
rhpijnacker: I updated it as follows:
image

Issue #173, opened 2024-11-25
🪲 BUG: Amber price data does not get sent


ArdJonker:
In the call to FMClient.post_measurements unit is used in stead of uom.

PR #174, opened 2024-11-25
#173 🪲 BUG-FIX release 0.4.2: Amber price data does not get sent

ArdJonker:
- Fixed unit_of_measure problem for Amber
- Made fix for both Amber and Octopus that was introduced by refactoring for version 0.4.0.


ArdJonker (commented):
Self approved

PR #177, opened 2024-11-27
Add "Restart" button to ping/pong error message

rhpijnacker:
image

Pressing this "Restart" button will (re)start the V2G Liberty add-on (provided it is not a local installation).


v2g-liberty-cards/src/strings.json
ArdJonker, line 1: Now the error-sub text can be changed to:
A restart of the V2G Liberty add-on is needed, please click RESTART below.

rhpijnacker: Updated as requested.

PR #183, opened 2024-12-11
179 bug detection of charger crash not 100 accurate

ArdJonker:
Also added a sensor current_scheduled_charging_power representing the the power that is suggested by the schedule.
The added binary is_charge_power_different_from_schedule monitors if the real_charging_power deviates > 500W from the scheduled power. This in a later stage can be used in Home Assistant (automation) to monitor the charging.


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
```diff
@@ -786,7 +791,7 @@ async def set_records_in_chart(self, chart_line_name: ChartLine, records: dict):
     ######################################################################
 
     async def __pong(self, event, data, kwargs):
-        self.hass.fire_event("ping.result")
+        await self.hass.fire_event("ping.result")
```
rhpijnacker, line 794: Since this is "fire and forget", there is no real need to await this call.

v2g-liberty/rootfs/root/homeassistant/packages/v2g_liberty/v2g_liberty_package.yaml
```diff
@@ -242,8 +242,14 @@ input_number:
...
-    unit_of_measurement: "hrs"
-
+    unit_of_measurement: hrs
+
+template:
+  - binary_sensor:
+      - name: is_charge_power_different_from_schedule
+        device_class: problem
+        state: >
+          {{ ((states('sensor.current_scheduled_charging_power')|int - states('sensor.charger_real_charging_power')|int)|abs) > 500 }}
```
rhpijnacker, line 252: This does not have to be a problem perse, does it?
It could also mean that the load balancer is doing a good job.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -143,6 +157,12 @@ class ModbusEVSEclient:
         "ha_entity_name": "serial_number_low",
     }
 
+    # Groups of entities for efficient reading of the modbus registers.
+    CHARGER_POLLING_ENTITIES: list
+    CHARGER_ERROR_ENTITIES: list
+    # To be read once at init
```
rhpijnacker, line 163: Suggestion: instead of suggesting a usage in the comment, would it not make more sense to state that these entities contain "static" data, that only needs to be initialised once?
When exactly that happens seems like an implementation detail that does not belong in a comment.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -337,7 +366,10 @@ async def start_charge_with_power(self, kwargs: dict, *args, **fnc_kwargs):
             )
             return
 
-        charge_power_in_watt = int(float(kwargs["charge_power"]))
+        charge_power_in_watt = parse_to_int(kwargs.get("charge_power", None), None)
+        if charge_power_in_watt is None:
+            self.__log(f"invalid charge_power: {kwargs.get('charge_power', None)}")
```
rhpijnacker, line 371: Suggestion: put the result of `kwargs.get('charge_power', None)` in a local variable, so it does not have to be repeated.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -803,80 +838,131 @@ async def __get_and_process_registers(self, entities: list):
...
-            # call handler if defined
-            if "change_handler" in entity.keys():
-                str_action = entity["change_handler"]
+    async def __update_ha_and_evse_entity(self, evse_entity, new_value):
+        """Helper method to update both the evse- and ha-entity at the same time"""
+        await self.__update_evse_entity(evse_entity=evse_entity, new_value=new_value)
+        entity_id = f"sensor.{evse_entity['ha_entity_name']}"
+        await self.__update_ha_entity(entity_id=entity_id, new_value=new_value)
+
+    async def __update_evse_entity(self, evse_entity: dict, new_value):
```
rhpijnacker, line 895: The name of this method confuses me, as it does not write a value to the EVSE, or am I missing something?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -803,80 +838,131 @@ async def __get_and_process_registers(self, entities: list):
...
+              Defaults to None, can be "unknown".
+            attributes (dict, optional): The dict the attributes should be written
+            with. Defaults to {}.
         """
-        if state is None:
-            self.__log("called with state is None, aborting.")
-            return
-
-        new_attributes = None
+        new_attributes = {}
```
rhpijnacker, line 944: This function is a lot more complicated than strictly needed by the handling of the attributes. This is used only once. Maybe we are better of with a special function for that single use-case?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -803,80 +838,131 @@ async def __get_and_process_registers(self, entities: list):
...
         if self.hass.entity_exists(entity_id):
-            current_attributes = await self.hass.get_state(entity_id, attribute="all")
-            if current_attributes is not None:
-                new_attributes = current_attributes["attributes"]
-                if attributes is not None:
-                    new_attributes.update(attributes)
+            entity_state = await self.hass.get_state(entity_id, attribute="all")
+            # self.__log(f"{entity_state=}")
```
rhpijnacker, line 947: I suggest to do yourself a favor and remove all the "junk" so the code becomes more readable.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -1001,17 +1087,12 @@ async def __cancel_polling(self, reason: str = ""):
...
-        self.__log(f"__cancel_polling, reason: {reason}")
-        if self.hass.timer_running(self.poll_timer_handle):
-            await self.hass.cancel_timer(self.poll_timer_handle, True)
-            # To be really sure...
-            self.poll_timer_handle = None
-        else:
-            self.__log("__cancel_polling: No timer to cancel")
+        self.__log(f"reason: {reason}")
+        self.__cancel_timer(self.poll_timer_handle)
+        # To be really sure...
```
rhpijnacker, line 1094: Setting a timer to None after canceling it seems like common sense. So, I think adding this comment is very redundant.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -1042,90 +1123,121 @@ async def __base_polling(self, kwargs):
...
+        stored value from polling.
+        It is expected to be between min_value_at_forced_get/max_value_at_forced_get.
+        Keep trying to get a reading that is within min_value_at_forced_get - max_value_at_forced_get range or
+        a timeout has been reached.
+        After the timeout the value is checked against other min/max values. This is
+        a wider range.
+        This is aimed at the SoC, this is expected to be between 2 and 97%, but at
+        timeout 1% to 100% is acceptable.
+
+        If the value is not in the wider acceptable range at timeout we assume
```
rhpijnacker, line 1154: I can only assume that this is based on real-life experience. However, even the comment seems awefully complicated to me, which suggests the code will be also complicated, and therefore by extension have a high risk of complicated bugs.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
```diff
@@ -1042,90 +1123,121 @@ async def __base_polling(self, kwargs):
...
 
-        # If the real SoC is not available yet, keep trying for max. two minutes
+        # If the real SoC is not available yet, keep trying for
+        # max. self.MAX_CHARGER_ERROR_STATE_DURATION_IN_SECONDS
         while True:
             result = None
             try:
                 # Only one register is read so count = 1, the charger expects slave to be 1.
-                # I hate using the word 'slave', this should be 'server' but pyModbus has not changed this yet
+                # I hate using the word 'slave', this should be 'server' but pyModbus has
```
rhpijnacker, line 1177: Although I appeciate the sentiment, I hardly think this makes the code more readable.
You may want to consider refraining your personal opinions in code.
I do not expect this to ever change in pymodbus, as this is a simply a part of the modbus nomenclature, that is not going to change.

v2g-liberty/rootfs/root/homeassistant/packages/v2g_liberty/v2g_liberty_package.yaml
ArdJonker: True, it would not even have to be a problem if it lasts for a longer period (the intended use is to check this). On the other hand, the (only) alternative would be "generic", other device classes don't match at all. So device_class: problem is the best match. Maybe the threshold should be a user setting? 
Do you see a problem in using device_class: problem?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
ArdJonker: PyCharm (that i still use every now and then) gave a warning it should be awaited. But, I think you are right, I'll remove.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: 👍

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: 😔

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: It indeed is based on the (long) experience with the Quasar charger.
So, we need complex code to deal with some strange behaviour of the charger.
I agree that the comments are not helpful so i removed some.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: This is legacy of not being able to debug and relying on log statements for this. Wil remove.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: Yes, it writes:
    evse_entity["current_value"] = new_value
    evse_entity["previous_value"] = current_value

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
ArdJonker: 👍

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/modbus_evse_client.py
rhpijnacker: I would expect that set_state does that by default. If that is not the case, then I guess this will have to do.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
rhpijnacker: It does not really hurt to include it. So if that makes PyCharm happy ...

rhpijnacker (approved):

PR #186, opened 2024-12-17
181 bug cost data blinking at startup

ArdJonker:
Added "Loading..." to card content while relevant entities have no value.


ArdJonker (commented):
Self approved

PR #192, opened 2024-12-18
Dev

ArdJonker:
For version 0.4.3


v2g-liberty/config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: V2G Liberty
-version: 0.4.0
+version: 0.4.2
```
ArdJonker, line 3: Hmm, i expected 0.4.2 in Main as well. If installing from here i get version 0.4.2 anyway.

ArdJonker (commented):
All OK!

PR #197, opened 2024-12-19
#196 HOTFIX problems:

ArdJonker:
- input_text instead of sensor entities for chart lines
- disconnect detection not correct
- writing None values to sensors leads to HA errors


ArdJonker (commented):
Slef approved

Issue #203, opened 2025-01-07
Make connection status available for SmartSchedule card


ArdJonker:
- Include update of Look & Feel of all statuses (see Figma)
- Minor other updates on Look & Feel like button spacing, icon colors, etc

ArdJonker:
Fixed by #204

PR #204, opened 2025-01-07
#203

ArdJonker:
Added connection status to SmartSchedule card and set HA entity in fm_client. Made connection statuses more reliable/more frequent updates from fm_client, reservations_client and modbus_evse_client. Did some UI tweaks on cards and dialogs.


ArdJonker:
It has been tested on the test-instance outside of the devcontainer and number of columns are good there, could you check at your end?

v2g-liberty-cards/src/util/utils.ts
```diff
@@ -0,0 +1,14 @@
+export function elapsedTimeSince(dateTimeStamp: string) {
```
rhpijnacker, line 1: Suggestion: instead of creating one utils with a number of functions, consider creating smaller more dedicated files. E.g. here `time.ts` would be a good choice for all time related functions.

v2g-liberty-cards/src/car-reservation-calendar-settings-card.ts
```diff
@@ -90,6 +96,28 @@ export class CarReservationCalendarSettingsCard extends LitElement {
       : nothing;
   }
 
+
+  private _renderCaldavConnectionStatus() {
+    const state = this._caldavConnectionStatus.state;
+    const isConnected = state === 'Successfully connected';
```
rhpijnacker, line 102: In other places I created some enums for the string constants. We could consider doing the same here.

v2g-liberty-cards/src/car-reservation-calendar-settings-card.ts
```diff
@@ -90,6 +96,28 @@ export class CarReservationCalendarSettingsCard extends LitElement {
...
+    const hasConnectionError =
+      state === 'Connection error' || state === 'Failed to connect';
+    const error = tp('connection-error');
+    const success = tp('connection-success', {
+      time: elapsedTimeSince(this._caldavConnectionStatus.last_updated),
+    });
+    return isConnected
+      ? html`
+          <div class="success">
```
rhpijnacker, line 111: Consider using a `ha-alert alert-type="success"` instead. That keeps you from having to style it yourself.

v2g-liberty-cards/src/edit-charger-settings-dialog.ts
```diff
@@ -296,23 +297,33 @@ class EditChargerSettingsDialog extends DialogBase {
...
+        color: var(--error-color);
+      }
+
+      .name {
+        font-weight: bold;
+      }
+
+      // All of these don't seem to reach the hr element...
+      // hr,
+      // ha-markdown::part(hr),
```
rhpijnacker, line 315: It seems like a bad idea to me to want to style using css, instead of using markdown anyway ...

v2g-liberty-cards/src/schedule-settings-card.ts
```diff
@@ -88,4 +89,27 @@ export class ScheduleSettingsCard extends LitElement {
...
+    const isConnected = state === 'Successfully connected';
+    const hasConnectionError =
+      state === 'Connection error' || state === 'Failed to connect';
+    const error = tp('connection-error');
+    const success = tp('connection-success', {
+      time: elapsedTimeSince(this._fmConnectionStatus.last_updated),
+    });
+    return isConnected
+      ? html`
+          <div class="success">
```
rhpijnacker, line 106: Use `ha-alert` instead?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
```diff
@@ -67,18 +66,30 @@ def __init__(self, hass: Hass):
...
         )
 
-        # Ping every half hour. If offline a separate process will run to increase polling frequency.
+        # Ping every half hour. If offline, a separate process will run to increase frequency.
         self.connection_error_counter = 0
         # self.run_every(self.ping_server, "now", 30 * 60)
         self.handle_for_repeater = ""
 
     async def test_fm_connection(self, host_url, username, password):
+        """Test if charger can connect with given host and port
```
rhpijnacker, line 75: Something to consider: 
_Uncle Bob_ is a legend in the software development world, and author of the Clean Coding series of books/videos. He promotes writing as little comments as possible, as they are notoriously difficult to maintain. Instead, write your code so cleanly that you do not need the comment.
With this in mind, I wonder if you really need to add comments like this.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
```diff
@@ -341,12 +348,7 @@ async def __poll_dav_calendar(self, kwargs=None):
             )
             remote_v2g_events.append(v2g_event)
 
-        attributes = {"keep_alive": start}
-        await self.hass.set_state(
-            "sensor.calendar_account_connection_status",
-            state="Successfully connected",
-            attributes=attributes,
-        )
+        await self.v2g_main_app.set_caldav_connection_status(connected=True)
```
rhpijnacker, line 351: Replacing a event based notification with a direct method call to the v2g_main_app does not "loosen" the connection from reservations_client to v2g_liberty, which is highly desirable.
Instead, we should try to get reservations_client (and other modules) as unaware as possible from v2g_liberty.

v2g-liberty/rootfs/root/homeassistant/packages/v2g_liberty/settings_view.yaml
```diff
@@ -6,30 +6,32 @@
...
-        columns: 4
+      layout_options:
+        grid_columns: full
+
 - type: grid
   cards:
     - type: custom:v2g-liberty-car-settings-card
-      grid_options:
-        columns: 4
+      layout_options:
```
rhpijnacker, line 18: This appears to be working just fine. Nice!

v2g-liberty-cards/src/util/utils.ts
ArdJonker: OK, will do in next PR.

rhpijnacker: Why not update this PR instead?

v2g-liberty-cards/src/car-reservation-calendar-settings-card.ts
ArdJonker: Will do in next PR

v2g-liberty-cards/src/edit-charger-settings-dialog.ts
ArdJonker: Yes and no. The markdown `---` creates a <hr/> element that i would like to style more subtle than it is default.

v2g-liberty-cards/src/car-reservation-calendar-settings-card.ts
ArdJonker: No, in this case it renders not lake an alert but the box with a green line. This is not a standard HA element so keeping this HTML / style.

v2g-liberty-cards/src/edit-charger-settings-dialog.ts
ArdJonker: But leaving it for now as i don't get it to work.

v2g-liberty-cards/src/schedule-settings-card.ts
ArdJonker: No, in this case it renders not lake an alert but the box with a green line. This is not a standard HA element so keeping this HTML / style.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
ArdJonker: I agree with Uncle Bob here but the Linter complains... Can that just be ignored or can/should the linter be configured differently?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
ArdJonker: I agree. I was working towards an other goal: the reservations_client should not know about HA entities. The de-coupling will be fixed with the introduction of pyee.
What should get priority in your opinion?

v2g-liberty-cards/src/util/utils.ts
ArdJonker: Will do, what about utils_time.ts?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
rhpijnacker: We can certainly ignore it. And if we find it annoying, we should figure out how to silence the linter.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
rhpijnacker: Why do you think the reservation client should not know about HA entities? Especially the HA integration calendar is using HA entities anyway.

rhpijnacker: I think adding more calls back to other modules is a regression, so I would prefer replacing it with an event interface instead.

v2g-liberty-cards/src/util/utils_time.ts
```diff
@@ -0,0 +1,14 @@
+export function elapsedTimeSince(dateTimeStamp: string) {
```
rhpijnacker, line 1: Suggestion: rename to `time.ts` (as originally proposed).
The folder is already called `util/` so adding `utils_` another time is redundant.

v2g-liberty-cards/src/car-reservation-calendar-settings-card.ts
ArdJonker: OK, agree. I think the alert is a bit too 'alarming' but semantically it is the right thing. And I think the design should remain as much default HA as possible (so users can theme if they want). I'll change it.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
ArdJonker: OK, will keep it as is for now awaiting the pyee implementation.

v2g-liberty-cards/src/util/utils_time.ts
ArdJonker: Ah, oversaw that. Will do.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
ArdJonker: But in this case it also documents the somewhat complex returned values , so i think it still is a good thing to have a docstring here.
I do agree there are some docstrings just for the linter...

Issue #199, opened 2024-12-23
Make ping card look & feel complient with HA snackbar


ArdJonker:
The connection with back-end HA is communicated like this:
![afbeelding](https://github.com/user-attachments/assets/7dd8a2ca-5bf7-4182-8173-0dc8506e96d2)

As the ping card also refers to fail communication with a server it should comply with this look and feel.

Hint for code:
<img width="624" alt="afbeelding" src="https://github.com/user-attachments/assets/d4739c71-1ef0-49d3-83b7-d9be8abad849" />

PR #206, opened 2025-01-08
#199 Make ping card look & feel compliant with HA snackbar.

ArdJonker:
+ Uses mwc-snackbar now
+ Slightly different texts
+ Delay in start ping to prevent flickering
+ Delay in start ping after restart
+ Keep visible as long as the ping fails, even when restarting.
+ Moved to end of views to prevent the card taking up space even if not shown.


v2g-liberty-cards/src/ping-card.ts
```diff
@@ -1,10 +1,13 @@
-import { css, html, LitElement, nothing } from 'lit';
+import { html, LitElement, nothing } from 'lit';
 import { customElement, state } from 'lit/decorators';
 import { HomeAssistant, LovelaceCardConfig } from 'custom-card-helpers';
 
 import { callFunction } from './util/appdaemon';
 import { partial } from './util/translate';
 
+// Import the snackbar from MWC
+import '@material/mwc-snackbar';
```
rhpijnacker, line 9: This will increase the size of the generated file.
I would expect this component to be loaded already by HomeAssistant, so if we can avoid loading it here, that would save some space in the v2g-liberty-cards.js .

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -14,16 +17,19 @@ interface PingCardConfig {
 
 @customElement('v2g-liberty-ping-card')
 export class PingCard extends LitElement {
-  @state() private _isResponding: boolean;
+  @state() private _isResponding: boolean = true;
+  @state() private _snackbarOpen: boolean = false; // Snackbar open state
```
rhpijnacker, line 21: Suggestion: rename to  `_isSnackbarOpen`.
I propose to remove the comment, because the name is already clear enough.

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -14,16 +17,19 @@ interface PingCardConfig {
...
+  @state() private _isRestarting: boolean = false; // Track restart state
```
rhpijnacker, line 22: The variable is very clear, so I do not think the comment adds any value.

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -53,18 +63,22 @@ export class PingCard extends LitElement {
         this.hass,
         'ping',
         {},
-        this._config.ping_timeout * 1000
+        this._config.ping_timeout
       );
       this._isResponding = true;
+      this._snackbarOpen = false; // Close snackbar if ping is successful
```
rhpijnacker, line 69: I propose to remove the comments.

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -33,7 +39,11 @@ export class PingCard extends LitElement {
   public connectedCallback() {
     super.connectedCallback();
     this._connected = true;
-    this._startPinging();
+
+    // Add a delay to avoid snackbar flickering
```
rhpijnacker, line 43: This means an extra 15 seconds is added before the system detects that the addon is not responding. I do not think that is desirable.
I propose to fix the flickering in some other way.

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -79,30 +93,42 @@ export class PingCard extends LitElement {
...
+            persistent  <!-- Ensure it remains visible until manually closed -->
+            @closed=${this._onSnackbarClosed}
+          >
+            <mwc-button slot="action" @click=${this._restart}>${tp('restart')}</mwc-button>
+          </mwc-snackbar>
         `;
   }
 
-  async _restart() {
+  private _onSnackbarClosed() {
```
rhpijnacker, line 108: Suggestion: move this as an in-line definition in the method above:
```
  const _onSnackbarClose = () => this._isSnackbarOpen = false;
```

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -79,30 +93,42 @@ export class PingCard extends LitElement {
...
-  }
 
-  static styles = css`
-    .error {
-      font-weight: bold;
-    }
-  `;
+    // After the restart:
+    // - Add a delay before the next ping to allow for the restart to take effect
+    // - After the delay start pinging as usual
```
rhpijnacker, line 127: The `_ping` function already handles restarting the timer.
We do not want to start another sequence of timers, so I think we should remove this.

v2g-liberty-cards/src/ping-card.ts
```diff
@@ -79,30 +93,42 @@ export class PingCard extends LitElement {
...
-          </p>
-          <mwc-button @click=${this._restart}>${tp('restart')}</mwc-button>
+          <mwc-snackbar
+            ?open=${this._snackbarOpen}
+            labelText=${this._isRestarting ? tp('restarting') : tp('error')}
+            timeoutMs="-1"  <!-- Set timeout to -1 for persistent behavior -->
+            persistent  <!-- Ensure it remains visible until manually closed -->
+            @closed=${this._onSnackbarClosed}
+          >
+            <mwc-button slot="action" @click=${this._restart}>${tp('restart')}</mwc-button>
```
rhpijnacker, line 103: After hitting "Restart" the text changes to "Restarting..." but the button stays visible and can be used again.  I propose to remove the button while it is restarting, otherwise people may end up restarting the add-on multiple times.
Also I propose to change the restarting text to:
_Connection with V2G Liberty lost. Restarting..._

rhpijnacker (changes requested):

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Good call! The snackbar is available in HA for certain.
Would you have a suggestion on where/how to find out how to use that one?

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Will do

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Agree, will remove.

v2g-liberty-cards/src/ping-card.ts
ArdJonker: True. In production this seems no problem to me but i can image this is too long for dev.
Maybe start pinging and and suppress the error until the first ping is successful or if it takes longer than this timeout?

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Good suggestion, is more consistent.

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Good suggestion! Will do.

v2g-liberty-cards/src/ping-card.ts
ArdJonker: Ohh, i thought this was only influencing the frequency of the current timer, not starting a new one. This should prevent the error to overwrite the "restarting..." message during restart.
Will research this.

v2g-liberty-cards/src/ping-card.ts
rhpijnacker: I would expect that we can just use it like you do, without explicitly loading it ourselves.
Since HA already loads it, it should still work.

v2g-liberty-cards/src/ping-card.ts
ArdJonker: I now made the delay 1 second (instead of 15) and the unjustly showing of the error now is absent. A on second delay seems ok for dev as well.

ArdJonker (commented):
Self approved

PR #211, opened 2025-01-16
🔥 Hotfix on Main for failing local calendar

ArdJonker:
Adapted the detection of "all day" events and the __parse_to_tz_dt() method to be more robust against dates without time / time-zone info.


v2g-liberty/rootfs/root/homeassistant/www/v2g_liberty/v2g-liberty-cards/v2g-liberty-cards.js
rhpijnacker, line 1: It looks like you added these by accident. Are you sure you want to include these?

ArdJonker: No.. i did not want to include them but could not commit if i didn't.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/reservations_client.py
```diff
@@ -322,6 +328,14 @@ async def __poll_dav_calendar(self, kwargs=None):
         )
         await self.__process_v2g_events(remote_v2g_events)
 
+        attributes = {"keep_alive": start}
```
rhpijnacker, line 331: This looks like it is duplicated. I do not think that is what you intended.

ArdJonker (commented):
All OK.

PR #217, opened 2025-01-19
208 improve input validation on settingscards

ArdJonker:
Forgotten changes based on review comments that where in my stash... Oeps


ArdJonker (commented):
Approved

PR #220, opened 2025-01-29
Bumped pymodbus to 3.8.3. 

ArdJonker:
Tested on test environment with mock charger, will test on real charger.


ArdJonker (commented):
Self approved

PR #230, opened 2025-02-05
Build in extra safeguards not to accept a None or 0 value for the SoC and translate it to "unavailable".

ArdJonker:


ArdJonker (commented):
Self approved

PR #234, opened 2025-02-18
Fix addon build

rhpijnacker:
The Dockerfile installs the _build-base_ package so that it can use `patch`.
However, the functionality that is patched is not used by V2G Liberty.

I removed the patch file and the patching aswell as the build-base dependency.


ArdJonker (approved):
Tested locally, also outside devcontainer.

PR #248, opened 2025-03-12
247 bug when boosting for min soc and this is reached software hangs no new schedule is requested

ArdJonker:


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
```diff
@@ -333,7 +333,7 @@ async def set_next_action(self, v2g_args=None):
                 )
                 return
 
-            if soc > c.CAR_MIN_SOC_IN_PERCENT and self.in_boost_to_reach_min_soc:
+            if soc >= c.CAR_MIN_SOC_IN_PERCENT and self.in_boost_to_reach_min_soc:
```
rhpijnacker, line 338: This looks ok to me now.

rhpijnacker: The next check is on `soc <= (c.CAR_MIN_SOC_IN_PERCENT + 1)`.
Why is the `+ 1` needed?
Wouldn't that mean that it stops loading 1% too soon?

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_app.py
```diff
@@ -60,4 +60,4 @@ async def initialize(self):
         await octopus_price_data_manager.initialize()
 
         await v2g_globals.kick_off_settings()
-        await v2g_liberty.initialise_v2g_liberty(v2g_args="initialise")
+        await v2g_liberty.kickoff_v2g_liberty(v2g_args="initialise")
```
rhpijnacker, line 63: Do we want to stay consistent and use `kick_off_...` like on the previous line?

rhpijnacker (approved):

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
ArdJonker: Yes, this is to avoid problems with reaching the min-soc, i think the schedule should be requested with this extra "buffer" to avoid the problem.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_app.py
ArdJonker: true! Will do.

v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_liberty.py
ArdJonker: I'll make a new issue to get into the +1 stuff.

PR #249, opened 2025-03-13
Added feature to send version to fm

ArdJonker:
- Added set_version and __set_attribute to fm_client
- Added constant FM_ASSET_NAME, written by globals.
- Added (delayed) call to new method log_version to v2g_liberty

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/constants.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/constants.py
@@ -166,6 +166,9 @@ DEFAULT_UTILITY_CONTEXTS = {
 FM_ACCOUNT_USERNAME: str = ""
 FM_ACCOUNT_PASSWORD: str = ""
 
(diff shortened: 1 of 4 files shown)


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/constants.py
```diff
@@ -166,6 +166,9 @@
 FM_ACCOUNT_USERNAME: str = ""
 FM_ACCOUNT_PASSWORD: str = ""
 
+# Name of the FM asset that is managed, e.g. "John's Quasar"
+FM_ASSET_NAME: str = ""
```
rhpijnacker, line 170: How will this get the initial value?
What is the added value of adding this to V2G Liberty?

ArdJonker: The asset name is needed for retrieving the asset_id. That is needed for sending the asset attributes to FM.

PR #255, opened 2025-04-02
Pass the SOC sensor ID in the flex-model so that the soc-schedules get recorded in FlexMeasures

Flix6x:
This also means that you should be able to then replay the SOC schedules in the FlexMeasures UI, which will help with understanding schedules and/or debugging flex-models or scheduler implementations.

Please test.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/fm_client.py
@@ -653,6 +653,7 @@ class FMClient:
             "roundtrip-efficiency": c.ROUNDTRIP_EFFICIENCY_FACTOR,
             "consumption-capacity": max_consumption_power_ranges,
             "production-capacity": max_production_power_ranges,
+            "state-of-charge": {"sensor": c.FM_ACCOUNT_SOC_SENSOR_ID},
         }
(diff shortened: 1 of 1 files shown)


ArdJonker:
Tried this from within the devcontainer with devtest account and got this error:
ERROR:root:
        status: 422
        headers: <CIMultiDictProxy('Server': 'gunicorn', 'Date': 'Thu, 03 Apr 2025 19:26:47 GMT', 'Content-Type': 'application/json', 'Content-Length': '172', 'Vary': 'Cookie')>
        payload: {'message': {'json': {'state-of-charge': ['Cannot convert % to MWh']}}, 'result': 'Rejected', 'status': 'UNPROCESSABLE_ENTITY'}.

Flix6x:
Ah, too bad. We would be able to do the conversion in FM, but we'd need to make an assumption on the storage capacity. I don't think using the soc-max from the flex-model is a particularly robust approach, because that may be changing over time, but perhaps we should just try it out on our side. I'll keep this PR open and let you know when you could try again.

Btw I advise to set your default base branch to dev. That's probably a GH repository setting.

Flix6x:
This needs another try against FlexMeasures v0.32.0, which includes https://github.com/FlexMeasures/flexmeasures/pull/1996.

ArdJonker:
This change has been incorporated into #440 (adapted for the current codebase). Thank you @Flix6x!

PR #262, opened 2025-04-22
fix: stop using deprecated flex-context fields

Flix6x:
`consumption-price-sensor` and `production-price-sensor` are deprecated in favour of `consumption-price` and `production-price`.


ArdJonker:
It is fine if you squash-merge it into dev!

PR #280, opened 2025-06-05
265 refactoring use event bus instead of ha entities

ArdJonker:


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/event_bus.py
```diff
@@ -1,16 +1,30 @@
-from pyee import EventEmitter
+from pyee.asyncio import AsyncIOEventEmitter
 import log_wrapper
 from appdaemon.plugins.hass.hassapi import Hass
 
 
-class EventBus(EventEmitter):
+class EventBus(AsyncIOEventEmitter):
```
rhpijnacker, line 6: As discussed before, I am not convinced that having one shared event bus for all events is the right choice.

PR #284, opened 2025-06-11
Added `power-capacity` to flex_model in `fm_client`. 

ArdJonker:
Needs to be tested with FM setting lower/higher than local setting.


ArdJonker:
Need some help with testing. Case: FM setting differs from V2G Setting.

Flix6x (approved):
Yes, exactly. This becomes important to set when relaxing the consumption-capacity and production-capacity constraints.

PR #287, opened 2025-06-24
Removed "altered soc_minima" for 1st calendar_event.

ArdJonker:
Added `relax-soc-constraints` and `relax-capacity-constraints` to `flex_context`.


ArdJonker (commented):
Self approved

PR #302, opened 2025-07-03
Changed margin around SoC % element.

ArdJonker:
Removed obsolete `info_icon_style.yaml` and `input_number_style.yaml` files.


ArdJonker (commented):
Self approved

PR #308, opened 2025-08-20
Added chmod to Dockerfile

ArdJonker:


ArdJonker:
Problem was due to ineffective coping of files.

PR #315, opened 2025-09-17
314 feature request prepare for dynamic prices per 15 min

ArdJonker:


v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/v2g_app.py
```diff
@@ -72,3 +72,5 @@ async def initialize(self):
 
         await v2g_globals.kick_off_settings()
         await main_app.kick_off_v2g_liberty(v2g_args="initialise")
+
+        await get_fm_data.initialize()
```
rhpijnacker, line 76: Could this be above `kick_off_settings()`?
It looks a bit out of place, compared to the other `initialize` calls.

ArdJonker: True, it seems a bit out of place but the `get_fm_data.initialize()` method depends on global constants that are set by `v2g_globals.kick_off_settings()`.

PR #356, opened 2025-11-01
Improve load balancing

rhpijnacker:
This is an improvement to the load balancer to keep it from overshooting. It now takes the actual current power rate into account when setting the new power value, whereas previously it used a hard-coded value.
Because the Quasar is a bit slow in responding to new power values, this makes a huge difference.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/load_balancer/load_balancer.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/load_balancer/load_balancer.py
@@ -39,6 +39,7 @@ class LoadBalancer:
             return
 
         total_power = abs(total_power)
+        active_power = abs(self.rate_limiter.active_power)
         current_limit = self.rate_limiter.limit
 
         if total_power > self.total_power_upper_limit:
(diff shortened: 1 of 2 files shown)


v2g-liberty/rootfs/root/appdaemon/apps/load_balancer/load_balancer.py
```diff
@@ -92,21 +93,21 @@ def reset_low_timer(self):
 
     def reduce_power(self, kwargs):
         median_adjustment = int(statistics.median(self.high_values))
-        current_limit = self.rate_limiter.limit
-        new_limit = max(self.min_charge_power, current_limit - median_adjustment)
+        # current_limit = self.rate_limiter.limit
```
ArdJonker: Why not use the current limit any more?

ArdJonker (approved):
It is hard to judge the changes as i still have trouble getting the bigger picture of the working of the load-balancer. It seems very logical to to react to the active power instead of a pre-set limit.

v2g-liberty/rootfs/root/appdaemon/apps/load_balancer/load_balancer.py
rhpijnacker: The current limit is what we wish it to be.
However, the Quasar responds very slowly, so it is not necessarily there yet.
This is why I changed it to use the "active power" instead, which is (presumably) on the way to becoming the current_limit.

Using the actual active power avoids overshoots that were happening because of the slowly responding Quasar.

PR #364, opened 2025-12-23
Fix load balancer unit tests

rhpijnacker:
Forgot to update the unit tests after the latest load balancer changes.
The unit test was calculating with an uninitialized mock property.
Initializing the property properly fixes the tests.


ArdJonker (approved):
Viewed files, ran tests locally. Works fine again!

PR #381, opened 2026-01-19
Add hostname support for charger host configuration

ArdJonker:
- Renamed entity ID from charger_host_url to charger_host_name for clarity
- Updated charger settings card and edit dialog to use new entity ID
- Modified validation to accept both hostnames and IP addresses
- Updated UI strings and labels to reflect hostname support

Implements #380


ArdJonker (commented):
Self approved

PR #385, opened 2026-01-22
Fix domain exception during Notifier initialisation

ArdJonker:
- Moved _get_recipients() call from __init__ to async initialise() method to ensure Home Assistant domains are fully available before listing services
- Added try-except error handling around list_services() call to gracefully handle initialisation timing issues
- Added try-except around post_sticky_memo() to prevent crashes if HA not ready
- Changed "Configuration error" to "Configuration warning" for missing mobile devices
- Fixed formatting: spacing and line breaks per Ruff standards

This resolves intermittent domain exceptions that could occur during add-on startup when the Notifier tried to list services before Home Assistant was fully initialised.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- 🪲 BUG: Domain exception during Notifier initialisation causing startup failures (#385)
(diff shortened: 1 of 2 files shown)


ArdJonker (commented):
Self approved

PR #392, opened 2026-01-30
Improved forecast view

ArdJonker:
FlexMeasures generates forecasts for it's prices and adds these to the results when requesting new prices.
V2G Liberty now shows the forecasted prices as a dotted line in the chart.

A big refactor has been done to simplify the `get_flexmeasures_data.py` module and to make this code more testable.
Test coverage is now over 90% for all the code involved in getting prices and processing them.


ArdJonker (commented):
The `setup-devcontainer.sh` does not really belong in this PR as it was pulled earlier but it was needed for development on this work so it was added anyway.

Further: self approved

PR #394, opened 2026-01-31
Improved forecast view

ArdJonker:
Addition to PR #392 
Changed approach for getting data:
- First try get datetime of newest entsoe price.
- If that i s valid then get production & consumption price and emissions (not at a separate time any more).
Reduces number of calls to FM significantly.
Reduced number of code lines significantly.

Removed obsolete retry_handler.
- Calculate duration dynamically based on days_back
- Add logging for price fetch requests
- Optimise _find_latest_value_datetime to iterate backwards (data has trailing Nones, so this finds the last value quickly)


ArdJonker (commented):
Self approved

PR #397, opened 2026-02-04
Fix fm data importer initialise

ArdJonker:
Summary of the fix:

The bug was in fm_data_importer.py:240. When using AppDaemon's run_in() with keyword arguments like is_initial_call=True, AppDaemon passes them as a dictionary inside args[0], not as actual Python keyword arguments.

Before:

``` python
async def daily_kickoff_price_data(self, *args, is_initial_call: bool = False):
    # is_initial_call was always False because it was stuck in args[0]
```
After:

``` python
async def daily_kickoff_price_data(self, *args):
    # Extract from AppDaemon's callback dict
    is_initial_call = False
    if args and isinstance(args[0], dict):
        is_initial_call = args[0].get("is_initial_call", False)
```

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -1,23 +1,14 @@
 # What's changed?
 
-## 0.7.3 2026-01-31
+## 0.7.4 2026-02-04
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


ArdJonker (commented):
Self approved

PR #399, opened 2026-02-13
Add VSCode tasks for v2g-liberty-cards workspace

ArdJonker:
- Add NPM Build task to compile cards with Parcel
- Add NPM Watch task for live development with auto-rebuild
- Provides quick access to common development commands

Much to my surprise this only was a local file, not in github.


ArdJonker (commented):
Self review

PR #400, opened 2026-02-15
371 bug charge and discharge function sometimes slow

ArdJonker:
Forgot to add this to 0.7.3/4 releases.
Changelog admin to prepare version 0.7.5


ArdJonker (commented):
Self approved

PR #403, opened 2026-02-19
Fixed gap in chartlines fixed and forecast for consumption, productio…

ArdJonker:
…n and emissions.


ArdJonker (commented):
Self approved

PR #404, opened 2026-02-19
Auto-configure configuration.yaml with V2G Liberty package include on startup

ArdJonker:
Previously users had to manually edit configuration.yaml to add the V2G Liberty package include line. This adds a startup script that automatically detects and inserts the required line, creating a backup before modifying the file. Handles edge cases such as missing homeassistant: section, !include_dir packages, and !include delegation. Sends HA persistent notifications to inform the user of changes or when manual action is needed.

DOCS.md updated to reflect the new auto-configuration flow, with manual steps moved to a collapsible fallback section.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Added
 
+- 🚀 FEAT: Auto-configure configuration.yaml (#404)
+
 ### Changed
 
 &nbsp;
--- a/v2g-liberty/changelog_of_all_releases.md
+++ b/v2g-liberty/changelog_of_all_releases.md
@@ -12,6 +12,8 @@ That file also contains possible changes that the next release might include.
 
 ### Added
 
+- 🚀 FEAT: Auto-configure configuration.yaml (#404)
+
 ### Changed
 
 ## 0.7.5 2026-02-15


v2g-liberty/rootfs/root/scripts/ensure_ha_package_include.py
```diff
@@ -0,0 +1,302 @@
...
+
+    ha_line_idx = find_homeassistant_line(lines)
+
+    if ha_line_idx is None:
+        # No homeassistant: section -- prepend the entire block
+        block = "homeassistant:\n  packages:\n" + V2G_PACK_LINE + "\n"
+        # Ensure there is a blank line separating from existing content
+        if content.strip():
+            block += "\n"
+        return (
```
rhpijnacker: Writing functions with multiple return statements is generally considered a "code smell". Consider rewriting this.

v2g-liberty/rootfs/root/scripts/ensure_ha_package_include.py
```diff
@@ -0,0 +1,302 @@
...
+                "packages/v2g_liberty/v2g_liberty_package.yaml\n"
+                "```\n\n"
+                "Then restart Home Assistant."
+            ),
+            notification_id="v2g_config_yaml_manual",
+        )
+        return 0
+
+    # Cases that require modification: create backup first
+    backup_path = create_backup(content)
```
rhpijnacker: Does this create a new backup _every time_ we restart V2G-Liberty?
I regularly do that, so that means I will get a large number of backups.
I think we should include a clean-up mechanism to leave only a small number of backups, or even better: only backup when the content is actually modified.

ArdJonker:
I agree there is a risk but there also is value for (none developer) users. And I should do my best to minimise the risk.
Having said that, I think the chance that this script breaks the .yaml is smaller than a human breaking the .yaml?
How/what cases would i need to test in your opinion?

ArdJonker:
Would it be accaptable it the user would be asked by the system:
"I suggest to make these chances to the configuration.yaml:
x...
y...
z...
[Accept] [Reject]
A backup of the original configuration file will be made first. An invalid configuration file might be hard to recover from. Please check the above file thoroughly."

Would this give you enough certainty?

PR #409, opened 2026-03-05
🪲 BUG: Fix crash on partial Modbus response

ArdJonker:
Add bounds check in __get_and_process_registers to handle cases where the charger returns fewer registers than expected. Previously this caused an IndexError. Now partial responses are logged and escalated through the existing unrecoverable error handling.


ArdJonker (commented):
Self approved.
Hard to test, has run for a while but the situation where length og results < expected has not been mimicked.

PR #410, opened 2026-03-05
BUG: Fix schedule request flooding when previous request is lost

ArdJonker:
When a schedule request was detected as "lost" (taking too long), fm_date_time_last_schedule was not updated. This caused every subsequent call to get_new_schedule (triggered every few seconds by polling) to also consider the new request as "lost", firing 10-15+ concurrent requests to FlexMeasures. This resulted in slow scheduling and the StorageFallbackScheduler being used instead of the optimal one.

Also truncate the flex_model log message to 1500 chars to prevent oversized log entries from being dropped by the logging system.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


ArdJonker (commented):
Self approved

PR #414, opened 2026-03-09
BUG: Restore anti-flash guard in settings-error-alert-card

ArdJonker:
The _hasUninitialisedEntities guard, requestAnimationFrame deferral, and location-change listener were lost when the branch was created from a stale base. Without the guard, the dialog was shown on every hass update, causing a continuous open/close flash.


ArdJonker (commented):
Self approved

PR #444, opened 2026-05-09
🪲 BUG: Fix orphaned schedule timers overriding charge mode changes

ArdJonker:
Cancel existing charging timers before creating new ones in __process_schedule to prevent orphaned timers when two schedule responses arrive near-simultaneously. Previously, timers were created first and cancelled after, allowing a race condition where one set of timer handles could be lost while the timers themselves remained live in AppDaemon.

Remove __reset_charging_timers as it is no longer needed.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


ArdJonker (commented):
Self approved

PR #445, opened 2026-05-11
Refactor: make AppDaemon timer-API usage consistently async in main_app

ArdJonker:
- Add `set_at_timer` helper to timer_utils.py (cancel-previous + run_at in one call, matching the existing `set_oneshot_timer` pattern)
- Replace the last raw `hass.cancel_timer` call with `timer_running` + `cancel_timer_silent`, guarding against None handles
- Refactor reservation start/end timers to use `set_at_timer`
- Fix type annotation: `List[AsyncGenerator]` → `List[str]` (handles are string UUIDs after await), remove unused AsyncGenerator import

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


ArdJonker (commented):
Self approved

PR #446, opened 2026-05-11
Reduce excessive FM schedule requests on SoC changes

ArdJonker:
- Only trigger a schedule refresh when SoC changed >= 2%-points since the last refresh (tracked via soc_at_last_schedule_refresh). None, non-numeric values, and first-run are always treated as significant. The 15-min watchdog timer still ensures periodic schedule refreshes.
- Remove misleading "Slow async listener" warning from EventBus: it measured wall clock time of create_task'd coroutines, not actual event loop blocking, producing false alarms on normal async I/O waits.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


ArdJonker (commented):
Self approved

PR #447, opened 2026-05-11
Improve CBS fetcher error logging: include response snippet on JSON parse failure

ArdJonker:
When the CBS OData API returns non-JSON (e.g. an HTML redirect during an outage), the error log now includes the first 200 characters of the response body for easier diagnosis.


ArdJonker (commented):
Self approved

PR #449, opened 2026-05-15
Exclude __pycache__ from AppDaemon app discovery

ArdJonker:
To prevent flooding of the logs with
WARNING AppDaemon: Excessive time spent in utility loop: 2.8s, 2.8s in check_app_updates(), 7.988ms in other


ArdJonker (commented):
Self approved

PR #452, opened 2026-06-09
🪲 BUG: Fix broken UI (ha-xyz elements) on HA 2026.5+

ArdJonker:
### Background
Home Assistant's WebAwesome migration (2026.5+) removed several MDC components the V2G Liberty cards still relied on. As a result the settings dialogs rendered empty input fields, radio choices disappeared, and some loading spinners no longer showed. This PR brings the cards in line with the new components.

### What changed

**Inputs (`ha-textfield` → `ha-input`)**
- Added an entity-free `renderHaInput` primitive as the single source of truth for the input element.
- Migrated the charger, schedule/FM, electricity-contract and calendar dialogs (via the shared render helpers), plus grid, solar and reset-database.
- Number inputs suppress spin buttons by default; guarded `checkValidity` calls so a missing/late element can no longer crash the charger dialog.

**Radio choices (`ha-radio`/`ha-formfield` removed)**
- `ha-radio` was removed and its replacement `ha-radio-option` is lazy-loaded (it would reintroduce the empty-dialog problem), so radios are now drawn with a small CSS indicator using HA theme variables — works in light and dark mode and carries no lazy-loading risk.
- Settings selects render as a left-aligned radio list; the calendar source choice and the reset-database mode choice as radio cards (title + description).
- Restored the radio affordance on the grid / charger / solar phase pickers.
- **Fixes a real bug:** the reset-database mode (re-import / full) could not be selected at all on 2026.5+.

**Other removed/deprecated elements**
- `ha-circular-progress` → `ha-spinner` on the grid and solar loading cards.
- Updated the deprecated `ha-button size="small"` to `size="s"`.
- Kept the save spinner aligned within the new `wa-dialog` footer.

**Compatibility**
- `config.yaml` now requires Home Assistant 2026.5.0+, so the Supervisor blocks updating on older cores and shows a message.


ArdJonker (commented):
Self approved

PR #453, opened 2026-06-10
Fix perpetual SQLite lock and accept FlexMeasures HTTP 202

ArdJonker:
### Problem (observed in production on 0.8.1)
Two independent issues stopped the new grid/PV monitoring data from reaching FlexMeasures:

1. **Perpetual "database is locked".** The data repairer could run concurrently/repeatedly — the post-import full repair colliding with the API-triggered run (`v2g_run_full_repair`) and the 6-hourly incremental repair. Each rewrites the whole `interval_log` (~112k rows) in one transaction, and overlapping runs held the SQLite write lock long enough to **permanently** starve the live 5-minute monitoring writes (`data_monitor._write_interval_to_db` failed every 5 min for 14+ hours). As a result `interval_log`/`grid_interval_log`/`pv_interval_log` stopped being written, so there was nothing to send.

2. **HTTP 202 treated as a failure.** FlexMeasures returns `202 Accepted` for sensor-data posts, but `flexmeasures-client` 0.8.1 raises `ValueError("Request failed with status code 202")` (it whitelists a single status). `fm_client.post_sensor_data` treated this as a failure, so the data was retried forever and FM was flagged as down — even though the data had been accepted.

### Fixes

- **Single-flight guard** in `data_repairer`: only one repair (full or incremental) may run at a time; overlapping triggers are skipped. This addresses the repeated/overlapping repairs that caused the lock.
- **`PRAGMA busy_timeout = 30000`** on the shared `DataStore` connection as a backstop against short-lived contention.
- **`post_sensor_data`** now treats any HTTP 2xx (including 202) as success instead of a failure.
- Tests for the 202 handling (202/200/error) and the single-flight repair guard.

### Verification in production
After deploying these fixes and upgrading:
- No more "database is locked"; `grid_interval_log`/`pv_interval_log`/`interval_log` are being written again.
- `fm_send_status` for charger/grid/pv advanced past the frozen init timestamp — data is now accepted by FlexMeasures.

### Notes
- The 112k-row repair is no longer chunked into smaller transactions; the single-flight guard + `busy_timeout` proved sufficient in production, so chunking was intentionally left out to keep the change focused.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- 🪲 BUG: Fix perpetual database is lock + Accept FM HTTP 202 when posting sensor data (#453)
 - 🪲 BUG: Fix broken UI (ha-xyz elements) on HA 2026.5+ (#452)
 - 🪲 BUG: Reduce excessive FM schedule requests on SoC changes (#446)
 - 🪲 BUG: Fix orphaned schedule timers overriding charge mode changes (#444)
--- a/v2g-liberty/changelog_of_all_releases.md
+++ b/v2g-liberty/changelog_of_all_releases.md
@@ -7,6 +7,7 @@ That file also contains possible changes that the next release might include.
 
 ### Fixed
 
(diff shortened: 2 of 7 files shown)


ArdJonker (commented):
Self approved

PR #455, opened 2026-06-12
Clean prod image

ArdJonker:
### Background
The add-on image is built with `COPY rootfs /`, which copied **dev-only files into the shipped add-on**: the `dev_tools/` grid/PV emulator, both test suites (`appdaemon/tests/` and `/root/tests/`), and the dev AppDaemon config (`appdaemon.devcontainer.yaml`). They were never *loaded* (excluded from app discovery via `appdaemon.yaml` `exclude_dirs`), but they were still present in every installation. On top of that, the add-on's startup copy used `cp -aRv …`, logging **every single file** to the add-on log (~120 lines) on each start.

### Changes
- **Dockerfile** — after `COPY rootfs /`, strip the dev-only paths from the image:
  `apps/dev_tools/`, `appdaemon/tests/`, `appdaemon.devcontainer.yaml`, `/root/tests/`.
- **`init-appdaemon/run`** — copy config files **without `-v`** (no per-file flood) and emit a short `bashio::log.info` confirmation per step plus a final *"V2G Liberty configuration complete."* line. Copy failures are still reported via `bashio::exit.nok`.
- **`dev_tools/README.md`** — document that the emulator is stripped from the built image, and how to run it outside the dev container (run it in the dev container, or add it manually on a host).


ArdJonker (commented):
Self approved

PR #457, opened 2026-06-12
Forgotten remove of files

ArdJonker:
A bit messy, forgotten changes to run script need to go into the git release chain again..
- Remove old files copied by earlier releases that have no value. 
- Removed obsolete rename of v2g-liberty folder


ArdJonker (commented):
Self approved

PR #462, opened 2026-06-12
Dev

ArdJonker:
More errors to be fixed. Remove `- dev_tool` from `exclude` list in `appdaemon.yaml`: not needed as files are not copied there any more.


ArdJonker (commented):
Self approved

PR #464, opened 2026-06-17
Handle float values in total power sensor

sin7ek:
Some Home Assistant power sensors expose values as floating point strings (e.g. -1473.99997711182). > incoming power not int: new='-1473.99997711182'

The current implementation uses:

total_power = int(new)

which raises a ValueError and causes valid power updates to be ignored.

This change converts the value through float() first and rounds it to an integer:

total_power = int(round(float(new)))

This allows the load balancer to work with both integer and floating point power sensors while preserving the existing integer-based logic.

### DIFF
--- a/repository.yaml
+++ b/repository.yaml
@@ -1,3 +1,3 @@
-name: V2G Liberty - Automated and optimised vehicle-to-grid / bidirectional charging of your electric vehicle
+name: V2G Liberty - sin7ek - Automated and optimised vehicle-to-grid / bidirectional charging of your electric vehicle
 url: https://github.com/sin7ek/addon-v2g-liberty
 maintainer: "sin7ek <flexmeasures@hagenback.nl>"
--- a/repository.yaml
+++ b/repository.yaml
@@ -1,3 +1,3 @@
 name: V2G Liberty - Automated and optimised vehicle-to-grid / bidirectional charging of your electric vehicle
-url: https://github.com/V2G-liberty/addon-v2g-liberty
-maintainer: "Ard Jonker <post@v2g-liberty.eu>"
+url: https://github.com/sin7ek/addon-v2g-liberty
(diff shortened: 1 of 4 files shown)


ArdJonker (commented):
Thanks @sin7ek for catching and fixing this! 🙏

You're spot on: float-valued power sensors (e.g. `-1473.99997711182`) break the load balancer — `int(new)` raises `ValueError` and the update gets dropped. The `quasar_load_balancer.py` change is a good fix and we'll bring it in.

Two things so we can land it cleanly:

1. **This PR also changes two files that are environment/fork-specific and can't go upstream:**
   - `data_import/fetchers/entsoe_fetcher.py` — `ENTSOE_SENSOR_ID` / `ENTSOE_SOURCE_ID` are your FlexMeasures account's IDs; merging these would point everyone else's ENTSOE import at the wrong sensors.
   - `repository.yaml` — name / url / maintainer point to your fork.

   These look like they slipped in from your local setup.

2. **Target branch:** we integrate changes via `dev` (PRs target `dev`; releases go `dev → main`), not directly into `main`.

To get your fix in quickly and with credit, we'll cherry-pick just the `quasar_load_balancer.py` change into a `dev`-targeted PR and credit you. One tiny simplification we'll apply: in Python 3, `round(float(new))` already returns an `int`, so the outer `int()` isn't needed — `total_power = round(float(new))`. Catching `(ValueError, TypeError)` and removing the duplicate `total_power = int(new)` line are nice touches too.

Thanks again — much appreciated! I'll link the follow-up PR here.

sin7ek:
Whoops for _This PR also changes two files that are environment/fork-specific and can't go upstream:_ that wasn't intended. I am playing around with a local FM server.. Pointing the sensors towards there, etc. Slipped in, my bad.. I am new at PR's and stuff 😅

PR #465, opened 2026-06-18
Handle float values from the total power sensor in the load balancer

ArdJonker:
Some HA power sensors report floats or float-strings (e.g. "-1473.99997711182") for the total power. The load balancer parsed the incoming value with int(), which raised a ValueError on such input and dropped the reading, so no rate limiting was applied.

Parse the value with round(float(new)) instead, rounding to whole watts, and broaden the except to (ValueError, TypeError) to also cope with None.

Based on #464 by @sin7ek.


ArdJonker (commented):
Self approved

PR #466, opened 2026-06-19
Refactor: rename FlexMeasures "Main Connection" asset to "Mains Connection"

ArdJonker:
"Mains" (the electrical supply) is the correct British-English term for the grid connection. Rename the FM asset name used in provisioning/lookup and the internal FM_MAINS_CONNECTION_ASSET_ID constant accordingly, plus the affected tests. The existing accounts' assets are renamed on the FlexMeasures side separately.


ArdJonker (commented):
Self approved

PR #469, opened 2026-06-23
Fix wrong range in "battery at max SoC" notification after restart

ArdJonker:
The "Car battery at max SoC" notification could fire on the first SoC reading after a restart, when the car was already at the max-SoC setting. At that point the car settings (battery capacity) may not have been loaded yet, so the range was computed with the default capacity and reported a wrong value (e.g. 110 km instead of 278 km).

Only send the notification on a genuine rise to the max-SoC setting while charging: require old_soc to be a valid numeric previous value and lower than new_soc. This skips the initial post-restart reading and guarantees the range is computed after the settings are loaded.

Add unit tests covering the genuine-rise, post-restart, already-at-max and not-charging cases, all derived from CAR_MAX_SOC_IN_PERCENT so they follow the setting rather than a hardcoded value.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- 🪲 BUG: "Battery at max SoC" notification reports the wrong range after a restart (#___)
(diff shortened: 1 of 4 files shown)


ArdJonker (commented):
Self approved

PR #470, opened 2026-07-19
Fix db schema validation

ArdJonker:
## Summary

A database whose schema does not match what the code expects brought the whole add-on down at startup. This PR makes V2G Liberty detect that, disable data features gracefully, and start normally — charging and driving keep working. On top of that, the data page now shows an in-context notice instead of a misleading € 0.00 of savings.

## Background

A tester running v0.8.1 could no longer start the app:
```log
    ERROR OperationalError: no such column: naive_power_w
    ERROR AppStartFailure: App 'v2g-app' failed to start
    ERROR   OperationalError: no such column: data_type
    WARNING data_store: Database schema version 4 is newer than expected 2.
```

**Root cause:** the database schema numbering was reset in March 2026 (the version had climbed to 4 and was set back to 1). The tester's database still carried number 4 from that old line, but is in fact missing columns (`interval_log.naive_power_w`, `fm_send_status.data_type`). The code saw `4 > 2`, concluded "newer than expected", skipped the migration, reported success — and then crashed on the first consumer hitting a missing column. `initialise()` trusted the version number blindly and never checked the actual schema.

## Changes

**Schema validation on startup** (`data_store.py`)
- New `__validate_schema()` reads the actual columns via `PRAGMA table_info` and compares them against `EXPECTED_SCHEMA`. On a mismatch it raises a `RuntimeError` with a readable summary of what is missing.
- Runs as the last step of `initialise()`, after any migration: a genuine v1 database migrates first and then validates cleanly.
- Unknown extra columns are not an error — a database from a newer version stays usable on older code.
- The existing safety net in `v2g_app` disables data features on failure, shows a notification pointing to "Reset history", and lets the app start fully.
- A note at `CURRENT_SCHEMA_VERSION` records the March 2026 renumbering (version 3/4 is actually older, not newer).

**More robust naive charging simulator** (`naive_charging_simulator.py`)
- `_run_batch()` and `_on_interval_concluded()` run detached / as an event-bus callback; their bodies are now wrapped in try/except so a failure no longer surfaces as an "unhandled exception in the event loop".

**In-context savings notice** (`v2g-liberty-cards`)
- The savings card shows "Savings data not available yet" when there is no naive data for the whole period, instead of a misleading € 0.00. Detection is on the raw rows (before the `?? 0` collapse). New en/nl string and styling; rebuilt bundle included.

## Testing

- New unit tests in `test_data_store.py`: incompatible schema (raises + `is_available` False), fresh database validates cleanly, migrated v1 database validates cleanly, unknown extra column allowed, and graceful degradation (`get_fm_last_sent` returns `None`).
- Full suite green (929 tests).
- Verified end-to-end against reconstructed databases: `ronald` (version 4 + missing columns), `missing-naive` (version 2 but columns dropped), `v1` (migrates), `extra-column` (tolerated). In every case: no `AppStartFailure`, no `no such column`, the app starts fully and the car stays active. The savings notice was confirmed visually on the data page.


ArdJonker (commented):
Self approved

PR #472, opened 2026-07-22
Warn negative grid power

ArdJonker:
## Summary

Warns the user when a selected grid sensor reports a **negative** value. Consumption and production sensors must be directional (one for power drawn from the grid, one for power fed back), each zero or positive. A negative value means a net/combined sensor was picked — for example a CT clamp that also sees the car feeding back — or simply the wrong sensor. Left unnoticed, these negatives flow straight through to the `Grid Consumption L{phase}` sensors in FlexMeasures.

## Background

A user upgraded and saw negative values on the `Grid Consumption L1` sensor in FlexMeasures. Root cause: the grid sensors had been pointed at bidirectional CT sensors (which include the car feeding back) instead of the grid meter (P1). V2G Liberty passed the raw signed value straight through — the whole pipeline (`_handle_grid_consumption_change` → `PowerTracker` → FM) has no sign handling, and nothing warned the user.

Chosen approach: **warn, do not clamp** — clamping would silently discard data for a genuinely net sensor, and the requirement is really that the user selects directional sensors.

## Changes

**Runtime warning** (`data_monitor.py`)
- When a grid consumption/production entity reports a negative value, log a WARNING and raise a persistent notification pointing the user to the grid connection settings. Throttled to once per direction+phase per session. Measured values are left untouched (no clamping).

**Config-time warning** (`edit-grid-connection-settings-dialog.ts`)
- During the dialog's inline entity validation, a selected sensor that reads negative gets a per-row warning icon plus an alert explaining that consumption/production must be positive and that a single net sensor is not supported. Aligned with the existing inline-validation architecture (client-side `state_changed` subscription); no dead backend plumbing. Rebuilt cards bundle included.

**Emulator test aid** (`grid_pv_emulator.py`, dev-only)
- A per-phase `Emulated Grid Net L{n}` sensor (import positive, export negative) and an `input_boolean.emulator_force_negative` toggle, so both warnings can be exercised without real hardware.

## Testing

- New unit tests: negative reading raises the notification, warns only once per channel, a positive value does not warn.
- Full suite green (932 tests), ruff clean.
- Verified end-to-end with the emulator: selecting `Emulated Grid Net L1` in the dialog shows the ⚠ icon + alert; after saving and (currently) a restart, the runtime persistent notification appears when the sensor reads negative.


ArdJonker (commented):
Self approved

PR #473, opened 2026-07-22
Live reregister grid listeners

ArdJonker:
## Summary

Grid connection settings changes now take effect immediately — no app restart needed. Previously, `DataMonitor` registered its grid listeners only once at startup, so after a user changed the grid sensors, the app kept monitoring the old entities (or none) until it was restarted.

## Background

`DataMonitor._setup_grid_listeners()` was only ever called from `initialize()`. Saving new grid connection settings updated the `c.GRID_*` constants and provisioned the FlexMeasures assets, but never told `DataMonitor` — so the newly selected sensors were not monitored, and any data sent to FlexMeasures kept coming from the old entities, until a restart.

This also affected the negative-value warning added in #472: the runtime persistent notification only fired after a restart for a newly configured wrong/net sensor. With this change it fires as soon as the new sensor reports a negative value.

## Changes

- **`event_bus.py`** — document a new `grid_settings_changed` event (emitted by `v2g_globals`).
- **`v2g_globals.py` / `v2g_app.py`** — pass the `event_bus` into `V2GLibertyGlobals`; emit `grid_settings_changed` after a successful grid settings save (not on the FM-error/blocked path).
- **`data_monitor.py`** — keep the `listen_state` handles; add `_teardown_grid_listeners()` (cancel the listeners, reset the trackers/scales, and re-arm the negative-value warning) and `_on_grid_settings_changed()` (tear down, then re-register on the new entities); subscribe to the event in `initialize()`.

Scope note: only the grid listeners are re-registered. The residential-load calculation reads `c.GRID_PHASES` / `c.SOLAR_PANELS` / `c.CHARGER_CONNECTED_TO_PHASE` live at interval conclusion, so those pick up changes automatically. PV listeners bind to `SOLAR_PANELS` (not grid entities) and are out of scope; re-registering them on a solar-panel settings change is a possible follow-up.

## Testing

- New unit tests: listener handles are stored on setup; a settings change cancels the old listeners and registers on the new entities; the negative-warning flags are cleared; and the save handler emits `grid_settings_changed` on success but not on the FM-error path.
- Full suite green (936 tests), ruff clean (the three pre-existing `F841` warnings in `v2g_app.py` are unrelated).
- Verified end-to-end in the dev app: after saving a changed grid sensor, the log shows `_on_grid_settings_changed > Grid settings changed, re-registering grid listeners.` and monitoring switches to the new sensor without a restart.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Added
 
 - 🚀 FEAT: Warn negative grid power (#472)
+- 🚀 FEAT: Live reregister grid listeners (#473)
 
 ### Changed
 
--- a/v2g-liberty/changelog_of_all_releases.md
+++ b/v2g-liberty/changelog_of_all_releases.md
@@ -14,6 +14,7 @@ That file also contains possible changes that the next release might include.
 ### Added
 
 - 🚀 FEAT: Warn negative grid power (#472)
+- 🚀 FEAT: Live reregister grid listeners (#473)
 
 ## 0.8.2 2026-06-19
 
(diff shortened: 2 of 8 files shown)


ArdJonker (commented):
Self approved

PR #474, opened 2026-07-22
Explain the 'charger phase not set' warning and keep it in sync

ArdJonker:
## Summary

Improves the "Charger phase not set" warning on the charger settings card: it now explains what happens next, and — crucially — it appears and clears in sync with the actual state instead of going stale until a page reload.

## Background

On a 3-phase grid the charger's phase must be known. When it is not, the charger settings card showed a bare `Charger phase not configured.` The message had two shortcomings:

1. **No guidance.** It didn't say that the phase is detected automatically the next time the car connects, nor what to do if it keeps showing.
2. **Stale state.** The card only reloaded its phase info on `save_charger_phase.result`. So the warning was wrong after:
   - a grid settings change (which clears the charger phase — because the phase attribution depends on the grid config), and
   - a successful phase detection (which fires `detect_charger_phase.result`, not a save event).
   In both cases a manual page reload was needed for the card to reflect reality.

## Changes

`charger-settings-card.ts`:
- Reword the warning with a bold title and a plain-language explanation: it is detected automatically the next time the car is connected, and if it keeps showing, open the charger settings to detect or set it manually.
- Reload the phase info on two more events so the warning stays in sync:
  - `save_grid_connection_settings.result` — appears when a grid change clears the phase.
  - `detect_charger_phase.result` — clears when detection (automatic or manual) sets the phase.

Rebuilt cards bundle included.

## Testing

Verified end-to-end in the dev app:
- Changing the grid sensors clears the charger phase → the warning appears immediately (no reload).
- Triggering automatic phase detection (car reconnect) that succeeds → the warning clears immediately (no reload).


ArdJonker (commented):
Self approved

PR #475, opened 2026-07-25
Add charger emulator dev tool + charger-driver regression net

ArdJonker:
## Summary

Adds a **charger emulator** that makes the static Wallbox Quasar Modbus mock respond to V2G Liberty, so charging, phase detection, SoC and error handling can be tested without real hardware — plus a **regression net** that pins the current charger-driver behaviour, and an interactive dev dashboard to drive it all.

Everything here is **dev-only** (the `dev_tools` apps are stripped from the production image). It prepares the ground for the upcoming `359` charger/EV migration: the emulator is the prerequisite for end-to-end charger testing, and the regression net is that migration's Fase 0 safety net.

## What's included

**Charger emulator** — `dev_tools/charger_emulator.py`, `charger_scenarios.py`
- Second Modbus client to the mock: mirrors V2G's setpoint onto the report registers (actual power, state), with an asymmetric power ramp (slow up, fast down) hard-clamped to the hardware max, and a SoC ramp with a hardware taper.
- Scenarios: `normal`, `reduced_max_power`, `wrong_fingerprint`, `error_state`, `internal_error`; plus connect/disconnect and a manual "Set SoC".
- Crash / no-connection via container control (`docker compose pause`/`stop`), documented in the README.
- A shared, pure-data scenario/profile module reused by both the emulator and pytest.

**Regression net** — `tests/v2g_liberty/test_modbus_evse_client_scenarios.py`
- 9 tests that drive the real `modbus_evse_client` with scenario register states and assert the observable contract (event_bus events + entity values): decode, power/SoC/state changes, connect/disconnect, the SoC-0 glitch, errors and polling. Pins current behaviour so the 359 refactor can be verified to preserve it.

**Interactive dev dashboard** — `homeassistant/packages/dev_tools/`
- A dev HA package with real input entities + a standalone Lovelace dashboard (controls, live charger status, grid/PV toggles, and a "Test checklist" tab). `grid_pv_emulator` now reads the real entities too (virtual fallback).

**Emulator unit tests** — `tests/dev_tools/` — 17 tests (mirror, ramp, taper, scenarios, SoC-resume, disjoint write-set).

## How to use
See `dev_tools/README.md`. In the dev container the emulator loads automatically; to activate the interactive dashboard, copy the two `homeassistant/packages/dev_tools/` files into your HA config, add the package include + dashboard registration and restart HA (details in the package header).

## Review notes
- `appdaemon.devcontainer.yaml` sets **`production_mode: True`** (disables AppDaemon hot-reload for this dev config) — decide whether to keep.
- `dev_tools/apps.yaml` default **`soc_speedup: 30`** (fast SoC ramp for testing).
- The HA dev package + dashboard are committed as **dev-tooling source only** — not activated for end users (a normal install has no `configuration.yaml` entry for them).

**Dev-only** stripped from the production image.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/appdaemon.devcontainer.yaml
+++ b/v2g-liberty/rootfs/root/appdaemon/appdaemon.devcontainer.yaml
@@ -21,7 +21,7 @@ appdaemon:
     - util
     - utils
 
-  production_mode: False
+  production_mode: True
   plugins:
     HASS:
       type: hass
@@ -33,6 +33,10 @@ logs:
     filename: /config/logs/v2g_liberty_main.log
   error_log:
     filename: /config/logs/v2g_liberty_error.log
+  # Dev-only: dedicated log for the charger emulator (dev_tools.charger_emulator).
+  charger_emulator_log:
+    name: charger_emulator
+    filename: /config/logs/charger_emulator.log
 
 http:
   url: http://127.0.0.1:5050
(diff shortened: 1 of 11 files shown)


ArdJonker (commented):
Self approved

PR #476, opened 2026-07-28
359 fase0 vangnet

ArdJonker:
## What

Expands the charger-driver regression net
(`test_modbus_evse_client_scenarios.py`) from 9 to 42 tests so the upcoming
359 charger refactor can be verified to preserve behaviour. Test-only — no
production code changes.

## Why

359 restructures the `modbus_evse_client` monolith into a `chargers/` package
(MBR/MCE dataclasses, EV split). Before that driver is touched, this PR pins
the current observable behaviour of the reliability-critical lifecycle paths
the base net previously patched away, so the refactor can be proven
behaviour-neutral. This is "Fase 0" (the safety net) of the staged 359 → dev
migration.

## What is now pinned

Every new test drives the real driver and asserts on the observable boundary
(event_bus events + cached entity values + the direct `v2g_main_app` /
`notifier` calls the driver makes instead of an event):

- SoC-refresh dance (real `__get_car_soc`): both branches, `force_emit`, and
  the 0/None → `unavailable` coercion
- Forced-read loop (`__force_get_register`): strict/relaxed accept,
  timeout → None + un-recoverable escalation
- Communication loss/recovery: grace-timer arming, recovery (comms-fault
  reset), un-recoverable, bad-config sticky memo, and the write-exception and
  repeat-exception branches
- Charger-error-state escalation (real `__handle_charger_error_state_change`):
  final-check timer, escalation, cancel-on-clear
- Partial/short modbus result: arm + abort
- Poll-strategy base (5 s) / minimal (15 s) selection + `evse_polled` stop
  semantics (`stop=True` = polling torn down, not "disconnected")
- Power-deviation >500 W flag (no event, no charger action)

A fault mode (`raise` / `short`) is added to the fake Modbus client to drive
the communication-loss paths.

## Verification

- Full suite green: **995 passed** (was 962).
- The new tests were adversarially reviewed for false-greens / wrong-pins /
  not-real-path defects; none survived. Three assertions were hardened and two
  coverage gaps (charger-error escalation, partial-read) were closed as a
  result.

## Notes

- Test-only; no changelog entry (non-functional, like tooling changes).
- A companion per-event contract checklist documents each event's kwargs, emit
  condition, keep/migrate decision, and exact consumers — the draaiboek for the
  Fase 2 charger port.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/tests/v2g_liberty/test_modbus_evse_client_scenarios.py
+++ b/v2g-liberty/rootfs/root/appdaemon/tests/v2g_liberty/test_modbus_evse_client_scenarios.py
@@ -16,14 +16,19 @@ from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
+from pymodbus.exceptions import ModbusException
 
 from apps.dev_tools.charger_scenarios import (
     REG_ACTUAL_POWER,
     REG_ERROR_1,
+    REG_LOCKED,
     REG_SOC,
     REG_STATE,
     STATE_CHARGING,
+    STATE_DISCHARGING,
     STATE_DISCONNECTED,
+    STATE_ERROR,
+    STATE_LOCKED,
     STATE_PAUSED,
     int16_to_uint16,
 )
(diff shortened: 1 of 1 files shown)


ArdJonker (commented):
Self approved

PR #477, opened 2026-07-30
Emulator manual test guide

ArdJonker:
### What

Rewrites the dev charger-emulator dashboard's test tab from a terse checklist
into an exact **Do / See** manual-test guide, split across cards, with proper tab
icons. Dev-tooling only — the `dev_tools` dashboard is stripped from the built
add-on image and never ships to users.

### Why

Reliable manual end-to-end testing against the emulator is a prerequisite for
executing the staged `359` migration reliably (each phase has to be verified by
hand as well as by the suite). The old "Test checklist" tab was too terse to test
from with confidence.

### Changes (`homeassistant/packages/dev_tools/emulator_dev_dashboard.yaml`)

- **Exact step-by-step**: every step is now `Do:` (which control / scenario /
  command) → `See:` (which sensors / log lines + the pass criterion), as its own
  bullets. Covers setup, happy path (idle, phase detection, charge, discharge,
  pause, max-SoC range notification), scenarios, comms and grid/PV.
- **Split into separate cards** (setup · happy path · scenarios · comms/grid/reset)
  so Home Assistant's masonry uses the page width instead of one wide card.
- **Tab icons**: `controller-classic-outline` for the Emulators (controls) tab,
  `book-open-variant` for the guide tab (renamed to "Test guide").

### Corrections baked in

- **`reduced_max_power`** now documents both real behaviours: **(A)** without
  re-configuring the charger, V2G still requests up to its current max but the
  emulator clamps delivery to ~3700 W; **(B)** after re-running the connection test,
  V2G reads register 514 = 3700 and requests ≤ 3700 itself.
- **`wrong_fingerprint`** is marked **not meaningful on dev yet** — there is no
  charger-type / signature validation until the 359 charger-type work lands; skip
  it for now.
- **Phase detection** notes the emulator circularity (it loads the
  already-configured phase `c.CHARGER_CONNECTED_TO_PHASE`, so detection only
  confirms the configured phase — not an independent test yet).

### DIFF
--- a/v2g-liberty/rootfs/root/homeassistant/packages/dev_tools/emulator_dev_dashboard.yaml
+++ b/v2g-liberty/rootfs/root/homeassistant/packages/dev_tools/emulator_dev_dashboard.yaml
@@ -85,30 +85,84 @@ views:
     cards:
       - type: markdown
         content: |
-          ## End-to-end test checklist
+          ## End-to-end test guide (manual)
 
-          **Setup**
-          - Keep the first tab open (live status + controls).
-          - Watch two logs: `charger_emulator.log` and `v2g_liberty_main.log`.
-          - Speed up the SoC: `soc_speedup: 30` in `dev_tools/apps.yaml` + restart
-            AppDaemon. Use the **Set SoC** slider to jump to a start value.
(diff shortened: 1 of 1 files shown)


ArdJonker (commented):
Self approved

PR #478, opened 2026-07-30
Introduce an ElectricVehicle abstraction and read car SoC from it

ArdJonker:
### What

Adds an `evs/` package with an `ElectricVehicle` that owns the connected car's
state of charge, and repoints `main_app`'s SoC / kWh / range reads onto it. The
charger monolith (`modbus_evse_client.py`) is **not touched** and remains the
sole SoC emitter. Behaviour-preserving.

### Why

This is the **EV-first** step of the staged migration of branch `359` (charger +
EV refactor) into `dev`. In 359 the EV abstraction is a *foundation* of the
charger layer (both charger drivers import `ElectricVehicle` and route SoC via
it). Introducing the EV before the charger port avoids a double charger-refactor
and lets the later charger port wire the new driver to an EV that already exists.
It also decouples car-state reads from the charger.

### Design decisions

- **Feed = the EV subscribes to `soc_change`** (a **sync, non-emitting** handler),
  so the charger stays the sole emitter and *zero lines of the reliability-critical
  charger hot-path change*. The Fase-0 regression net stays green by construction.
  Because a sync pyee listener runs inline during `emit`, `ev.soc` is refreshed
  before the async `soc_change` consumers (e.g. `main_app`) run in the same
  dispatch — no registration-order fragility.
- **Derivations read the same live constants as the charger**
  (`c.CAR_MAX_CAPACITY_IN_KWH` / `c.CAR_CONSUMPTION_WH_PER_KM`), so
  `soc_kwh` / `remaining_range_km` are bit-identical to the charger getters in
  **every** state, including the boot window before the car settings load.
- **Keeps dev's `"unavailable"` string sentinel** (not 359's `None`), so
  consumers' `EMPTY_STATES` guards keep firing identically.
- **Lean scope (complexity pushback):** no `BaseEV` ABC, no vehicle registry, no
  min/max-SoC, clamping, `max_range_km` or `initialise_ev` — only what dev needs
  now. Those migrate when their `c.CAR_*` readers migrate.

### Changes

- **New:** `evs/__init__.py`, `evs/electric_vehicle.py` (`soc`, `soc_kwh`,
  `remaining_range_km`, non-emitting `update_soc`).
- `v2g_app.py`: construct the EV before `main_app` and wire it in.
- `main_app.py`: repoint 6 reads — `get_car_soc()` → `ev.soc` (×3),
  `get_car_soc_kwh()` → `ev.soc_kwh` (×2), `get_car_remaining_range()` →
  `ev.remaining_range_km` (×1, inside `__handle_soc_change`).
- **Tests:** `test_electric_vehicle.py` (new, incl. a **differential parity test**
  that drives one SoC through both the EV and the real charger getters and asserts
  they agree); a charger-side derivation-parity test added to the Fase-0 net;
  `test_main_app_max_soc_notification.py` fixture repointed to the EV.

### Behaviour-neutrality

- Derivations are bit-identical (same formula, same live constants); the
  differential parity test pins EV == charger at runtime.
- Adversarially reviewed: one boot-window init-timing divergence was found and
  fixed (the EV now reads the live constants instead of a value that was `None`
  until init); everything else was confirmed benign.
- Full suite: **1003 passed** (ruff clean). No consumer sees a changed value/type.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/__init__.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/__init__.py
@@ -1,0 +1,1 @@
+"""Electric-vehicle abstraction package."""
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/electric_vehicle.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/electric_vehicle.py
@@ -1,0 +1,75 @@
+"""Electric vehicle (EV) abstraction.
+
+Holds the connected car's state of charge and exposes its energy content and
+range, so consumers can read the car from one place instead of from the charger.
+
+Fase 1 of the 359 migration (EV-abstraction before the charger refactor): the
+charger monolith stays the SOLE emitter and owner of the SoC getters. This
+object subscribes to the ``soc_change`` event and mirrors the value with a
+SYNC, NON-emitting setter, and derives soc_kwh / remaining-range from the SAME
+live car constants the charger reads (``c.CAR_MAX_CAPACITY_IN_KWH`` /
+``c.CAR_CONSUMPTION_WH_PER_KM``). Reading the live constants — rather than a
+stored copy — keeps the derived values bit-identical to the charger in EVERY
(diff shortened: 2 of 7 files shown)


ArdJonker (commented):
Self approved

PR #479, opened 2026-08-01
Introduce an ElectricVehicle abstraction and read car SoC from it

ArdJonker:
## Summary

Adds an `evs/` package with an `ElectricVehicle` that owns the connected car's
state of charge, and repoints `main_app`'s SoC / kWh / range reads onto it. The
charger monolith (`modbus_evse_client.py`) is **not touched** and stays the sole
SoC emitter. Behaviour-preserving — this is the EV-first step of the staged `359`
migration, decoupling car-state reads from the charger ahead of the charger port.

### Design

- **The EV subscribes to `soc_change`** with a **sync, non-emitting** handler, so
  the charger remains the sole emitter and *zero lines of the charger hot-path
  change* (the Fase-0 regression net stays green by construction). A sync pyee
  listener runs inline during `emit`, so `ev.soc` is fresh before the async
  `soc_change` consumers run in the same dispatch — no registration-order fragility.
- **Derivations read the same live constants as the charger**
  (`c.CAR_MAX_CAPACITY_IN_KWH` / `c.CAR_CONSUMPTION_WH_PER_KM`), so `soc_kwh` /
  `remaining_range_km` are bit-identical to the charger getters in every state,
  including the boot window before the car settings load.
- Keeps dev's `"unavailable"` string sentinel (not `None`) so `EMPTY_STATES`
  guards keep firing identically.
- **Lean scope:** no `BaseEV` ABC, registry, min/max-SoC, clamping or
  `initialise_ev` — only what dev needs now.

### Changes

- **New:** `evs/__init__.py`, `evs/electric_vehicle.py` (`soc`, `soc_kwh`,
  `remaining_range_km`, non-emitting `update_soc`).
- `v2g_app.py`: construct the EV before `main_app` and wire it in.
- `main_app.py`: repoint 6 reads — `get_car_soc()` → `ev.soc` (×3),
  `get_car_soc_kwh()` → `ev.soc_kwh` (×2), `get_car_remaining_range()` →
  `ev.remaining_range_km` (×1, inside `__handle_soc_change`).
- **Tests:** `test_electric_vehicle.py` (new, incl. a **differential parity test**
  that drives one SoC through both the EV and the real charger getters and asserts
  they agree); a charger-side derivation-parity test added to the Fase-0 net;
  `test_main_app_max_soc_notification.py` fixture repointed to the EV.

### Behaviour-neutrality & testing

- Derivations are bit-identical (same formula + same live constants); the
  differential parity test pins EV == charger at runtime.
- Adversarially reviewed: one boot-window init-timing divergence was found and
  fixed (the EV reads the live constants instead of a value that was `None` until
  init); everything else confirmed benign.
- Full suite: **1003 passed**, ruff clean. No consumer sees a changed value/type.

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/__init__.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/__init__.py
@@ -1,0 +1,1 @@
+"""Electric-vehicle abstraction package."""
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/electric_vehicle.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/evs/electric_vehicle.py
@@ -1,0 +1,75 @@
+"""Electric vehicle (EV) abstraction.
+
+Holds the connected car's state of charge and exposes its energy content and
+range, so consumers can read the car from one place instead of from the charger.
+
+Fase 1 of the 359 migration (EV-abstraction before the charger refactor): the
+charger monolith stays the SOLE emitter and owner of the SoC getters. This
+object subscribes to the ``soc_change`` event and mirrors the value with a
+SYNC, NON-emitting setter, and derives soc_kwh / remaining-range from the SAME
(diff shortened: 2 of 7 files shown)


ArdJonker (commented):
Self approved

PR #480, opened 2026-08-04
Auto-switch to Automatic when a paused reconnect prompt goes unanswered

ArdJonker:
## Problem (old bug)
When charge_mode is Pause (Stop), control is handed to the charger, which
re-enables its own autostart-on-connect. If the car is reconnected while the
app is left on Pause (e.g. forgotten), the charger autostart-charges the car to
full — defeating the "pause" intent. The existing reconnect notification asked
to switch to Automatic, but on ttl expiry it only cleared the message and took
no action, so a non-responding user's car still charged to full.

## Fix
- Lower the reconnect notification ttl from 30 → 10 min.
- If the user does not respond within that time, switch charge_mode to Automatic
  automatically, so the car comes under V2G control instead of the charger's
  autostart.
- Any deliberate response cancels the fallback: the notification buttons, a
  disconnect, or any charge_mode change via the UI radio buttons (new
  `input_select.charge_mode` listener registered in `initialize()`).

### DIFF
--- a/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/monitor_pause_at_reconnect.py
+++ b/v2g-liberty/rootfs/root/appdaemon/apps/v2g_liberty/monitor_pause_at_reconnect.py
@@ -2,10 +2,10 @@
 
 from appdaemon.plugins.hass.hassapi import Hass
 
-from . import constants as c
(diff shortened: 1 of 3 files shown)


ArdJonker (commented):
Self approved

PR #483, opened 2026-08-04
Guard the max-SoC notification against a non-numeric new SoC

ArdJonker:
## Problem
On car disconnect the charger reports SoC as `"unavailable"`, so
`__handle_soc_change` gets `new_soc="unavailable"` while `old_soc` is still a
number. The "battery reached max SoC" gate guarded `old_soc` but not `new_soc`,
so `old_soc < new_soc` raised `TypeError: '<' not supported between instances of
'int' and 'str'` on every disconnect. Caught by the event bus (no crash, no
functional harm), but a warning in the log each disconnect. Surfaced during the
fase2a emulator log review.

## Fix
Add the symmetric `isinstance(new_soc, (int, float))` guard beside the existing
`old_soc` guard.

## Testing
- New regression test (proven: fails with the exact TypeError without the fix).
- ruff clean on the changed lines; full suite 1033 passed.


ArdJonker (commented):
Self approved

PR #485, opened 2026-08-05
Fix grid connection save fm gate

ArdJonker:
## What & why

Saving the grid-connection setup provisions ~14 FlexMeasures sensors before it
persists, all-or-nothing. When FlexMeasures was slow or down, the frontend hit
its `callFunction` timeout and reported a useless "Failed to save" — nothing was
stored and the user got no actionable feedback. This PR makes that path fail
cleanly and self-heal, and reuses the same reachability gate on the solar dialog.

### Backend — `v2g_globals.py`
- Bound provisioning with `asyncio.wait_for` (both save and startup) so a slow or
  hung FlexMeasures fails cleanly instead of hanging past the frontend's timeout.
- Make startup provisioning non-blocking and self-heal once FlexMeasures
  reconnects (via the `fm_connection_status` listener), guarded against
  overlapping runs.
- Make provisioning atomic: publish the FM id constants only after every FM call
  succeeds, so an interrupted run no longer leaves a partial map that the
  self-heal guard mistakes for "already provisioned".
- Add a live reachability probe (`test_fm_reachable`) using the existing
  authenticated client, and log its outcome.

### Frontend — dialogs (`v2g-liberty-cards`)
- **Shared gate in `DialogBase`:** `_probeFm()` (a fresh live-client probe),
  `_fmReachable`, and `_renderFmGate(subject)` — a blocking error alert plus a
  "Check again" action — so any dialog whose Save provisions assets/sensors on
  the Smart schedule server can reuse one implementation.
- **Grid dialog:** gates step 1 on the fresh probe instead of the lagging
  `fm_connection_status` sensor; Continue stays disabled until FlexMeasures is
  reachable. The backend remains the authority at Save; clearer save errors.
- **Solar dialog:** probes on open and shows the shared gate as a takeover while
  the server is unreachable — covering both add and edit (edit skips the intro
  step, so an intro-only gate would miss it).
- **Charger dialog:** notes that once it starts provisioning/editing the charger
  asset on FlexMeasures (branch 359), it should apply the same shared gate.

## Tests
New `test_v2g_globals_grid_settings.py` covers the timeout, self-heal, atomicity,
listener wiring and the reachability probe.


ArdJonker (commented):
Self approved

PR #486, opened 2026-08-06
Keep the FlexMeasures connection green when a single sensor is rejected

ArdJonker:
## Summary

A single sensor's data post being rejected by FlexMeasures (e.g. HTTP 403 on a stale/foreign sensor id) no longer marks the **whole** FlexMeasures connection as `Error`. The connection status now reflects genuine connectivity only; an operational rejection of one post raises a persistent, actionable warning instead.

## Problem

`post_sensor_data` treated **every** non-2xx response as "connection lost" and called `set_fm_connection_status(connected=False)`. So one rejected sensor post — for example PV data sent to a stale sensor from an older FM account (HTTP 403 `FORBIDDEN`) — flipped the global status to `Error`, even though FlexMeasures was perfectly reachable (token OK, `get_assets` OK, grid posts OK). The sender retried every cycle, so the false `Error` never cleared on its own.

Two distinct concerns were conflated:

- **Connectivity / auth** — "can we reach FM at all?" (transport, timeout, 5xx, 401).
- **Operational rejection** — "FM is reachable but refuses this specific action" (403/404/422 on one sensor).

A `403` on one sensor is the second, but was reported as the first.

## Fix

`post_sensor_data` now classifies the exception:

- **Connection error** (no HTTP status → transport/timeout/DNS, or 5xx, or 401) → `set_fm_connection_status(connected=False)`, as before.
- **Operational 4xx** (403/404/422) → connection status is left untouched; a persistent HA memo ("FlexMeasures is refusing data…") is raised via a new `fm_data_issue` event, and dismissed automatically on the next successful post for that sensor.

Also fixes the 2xx-detection regex: the client raises errors in two shapes, and the previous `status code (\d{3})` pattern did not match the `Error occurred while communicating with the API: NNN` form — exactly the shape the observed 403 arrived in.

## Behaviour change

| Scenario | Before | After |
|---|---|---|
| 403/404/422 on one sensor's post | Whole FM connection shows `Error` (misleading, never clears) | Connection stays green; persistent actionable memo shown, self-clears on recovery |
| Transport/timeout/5xx/401 | Connection `Error` | Connection `Error` (unchanged) |

## Changes

- `fm_client.py` — `_fm_http_status` + `_is_connection_error` helpers; `post_sensor_data` classification + 2xx-regex fix; emits `fm_data_issue`; `_clear_data_issue` self-clear (tracks failing sensor ids so the memo stays up while any sensor is rejected).
- `main_app.py` — `handle_fm_data_issue` listener that shows/dismisses the sticky memo (mirrors the `no_new_schedule` pattern).
- `notifier_util.py` — `dismiss_sticky_memo` to clear a persistent notification by id.
- Tests — `test_fm_client_post_sensor_data.py`, `test_main_app_fm_data_issue.py`, plus additions to `test_notifier_util.py`.

## Testing

- Full suite: **1076 passed**.
- Live-verified against the dev environment (real FM staging + emulated PV): the 403 on the stale sensor is logged as `rejected (HTTP 403)`, `fm_connection_status` stays `Successfully connected`, and the persistent memo is shown in HA. Restoring a valid sensor dismisses the memo on the next successful post.

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -10,6 +10,7 @@
 - 🪲 BUG: Guard the max-SoC notification against a non-numeric new SoC (#483)
 - 🪲 BUG: Fix ttl-based notification clearing (unpack AppDaemon's kwargs dict) - (#484)
 - 🪲 BUG: Fix grid connection save fm gate - (#485)
+- 🪲 BUG: FlexMeasures connection wrongly shows "Error" when a single sensor's data is rejected (#486)
 
 
 ### Added
--- a/v2g-liberty/changelog_of_all_releases.md
+++ b/v2g-liberty/changelog_of_all_releases.md
@@ -13,6 +13,7 @@ That file also contains possible changes that the next release might include.
 - 🪲 BUG: Guard the max-SoC notification against a non-numeric new SoC (#483)
 - 🪲 BUG: Fix ttl-based notification clearing (unpack AppDaemon's kwargs dict) - (#484)
 - 🪲 BUG: Fix grid connection save fm gate - (#485)
+- 🪲 BUG: FlexMeasures connection wrongly shows "Error" when a single sensor's data is rejected (#486)
 
 
 ### Added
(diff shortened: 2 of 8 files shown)


ArdJonker (commented):
Self approved

PR #487, opened 2026-08-11
Feat aggregate energy dev

ArdJonker:
## Summary

Adds two FlexMeasures sensors on the **Mains Connection** — **Aggregate Consumption** and **Aggregate Production** — fed with the whole-connection import/export energy derived from the meter's **cumulative registers** (OBIS 1.8.x / 2.8.x). These are the exact, bill-reconcilable source for import and export.

This matters increasingly towards the **end of net metering (1 Jan 2027)**: once import and export are priced independently, only the metered registers give the correct per-interval energy. Per-phase power averages cannot reconstruct it — the meter's `max(·,0)` gate is applied continuously and per phase, so averaging power before the gate double-counts flows that never crossed the meter. The cumulative registers are gated inside the meter, so time-aggregation is loss-free.

## What's included

- **Detection** of the cumulative energy registers — attribute-based (`device_class=energy` + `state_class=total_increasing` + Wh/kWh/MWh), language-agnostic, with native-DSMR entity/translation-key patterns; sums tariff 1 + 2, or prefers a single total register.
- **Grid connection settings + UI** to select the import/export registers (2×2 table: tariff 1/2 × consumption/production). Selecting registers is **required** to save.
- **Per-interval delta** computation from the boundary register totals, with meter-reset and unavailable handling, plus baseline persistence across restarts.
- **Storage**: new `meter_interval_log` table (schema v3 → v4 migration).
- **FM provisioning** of the two kW-defined aggregate sensors.
- **Hourly send** of the per-interval kWh increments (posted as `kWh`; FM converts to average kW), tracked with a separate `meter` send cursor.
- Minor: log schedule-trigger failures with the full exception repr (status + URL); dev-only cumulative-register emulator (stripped from the production image by the Dockerfile).


ArdJonker (commented):
Self approved

PR #489, opened 2026-09-08
Bump FlexMeasures Client to 0.9.5

ArdJonker:
## Summary

Bumps `flexmeasures-client` from **0.8.1 → 0.9.5** in `requirements.txt`, for
**poll-based schedule retrieval**: the 0.9.5 client polls a still-computing
schedule (FlexMeasures signals "in progress" via HTTP 202 on older servers, or
HTTP 400 `STARTED` on FlexMeasures 1.0) through to `200` and a real schedule,
instead of treating the interim status as an error.

**No `fm_client.py` code change is required** — every client call V2G Liberty
uses stays backward-compatible in 0.9.5.

## What changed

- `requirements.txt`: `flexmeasures-client==0.8.1` → `0.9.5`

### DIFF
--- a/v2g-liberty/CHANGELOG.md
+++ b/v2g-liberty/CHANGELOG.md
@@ -24,6 +24,7 @@
 ### Changed
 
 - Explain the 'charger phase not set' warning and keep it in sync (#474)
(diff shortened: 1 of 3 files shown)


ArdJonker (commented):
Self approved

PR #491, opened 2026-09-11
evtec emulator profile

ArdJonker:
## Add an EVtec BiDiPro profile to the dev charger emulator

Completes task 27 of phase 3 of the 359 merg plan. The EVtec scenarios added earlier were not connected to anything; the emulator itself was Quasar-specific.

The charger-independent machinery (control entities, tick loop, power ramp, SoC integration, Modbus plumbing) moves to `charger_emulator_base.py`, with both emulators on top of it. The Quasar emulator keeps its observable behaviour and its existing tests pass unchanged.

### Deliberate choices, all from the hardware-tested contract

- **Suspend mode (X+88) is ignored** — it is a no-op on the real charger, so only a setpoint of 0 stops power flow.
- **Interlock violations deliver 0 W and log a warning** (session type X+04, or V2G not offered at X+22) instead of silently obeying. The real station does not guard this; making it visible is what lets tasks 26c/26d be tested.
- **X+22 advertises exactly what is delivered**, so the driver's clamp stays verifiable.
- **A setpoint outside the accept window is clamped but logged**, since the real station rejects it.

### Also included

An `evtec-mock` service (host port 5021), separate HA control entities so both emulators run side by side, a dedicated log and dashboard card. `setup-devcontainer.sh` now starts both mocks — it previously returned early once the Quasar mock was up, leaving `evtec-mock` down. The static mock JSON is generated from the emulator's own seed, with a test pinning the two together.

### Verification

49 dev_tools tests (was 38). Full suite: 1091 passed, with the pre-existing 66 errors unchanged. `loopback_check_evtec.py` checks the register contract against an in-process Modbus server using the driver's own read/write patterns: 21/21. That one is a script rather than a test, because no test in this repo binds a socket.

### Scope

Dev tooling only — every changed file is under `dev_tools/`, `charger-mocks/`, `packages/dev_tools/`, the compose file or the devcontainer config. Nothing in the production app, so no changelog entry (consistent with #475). End-to-end verification against the real driver has to wait for task 24.


ArdJonker (commented):
Self approved

PR #492, opened 2026-09-13
Keep the dev Home Assistant config out of the production example template

ArdJonker:
## What

`script/copy-config` overwrote `.devcontainer/config/configuration.yaml` with
`v2g_liberty_example_configuration.yaml` on every `setup-devcontainer.sh` run.
That example is the template shipped to users, so it cannot carry dev-only
entries — which meant the emulator package include was wiped each time and the
emulator control entities silently disappeared.

## Changes

- **`.devcontainer/init-config/configuration.yaml`** (new) — the dev environment
  gets its own template, holding the emulator package and its YAML-mode
  dashboard. `script/init-config` already seeds everything from `init-config/`,
  so a fresh environment picks it up too.
- **`script/copy-config`** — copies that template instead of the production
  example, which stays free of dev tooling.
- **`setup-devcontainer.sh`** — set the executable bit. It was committed as
  `100644`, so `./setup-devcontainer.sh` failed with "permission denied" on a
  fresh clone, while every other script in the repo is `100755`.


ArdJonker (commented):
Self approved

PR #493, opened 2026-09-13
Fix evtec mock input register

ArdJonker:
## What

`evtec-mock` started and exited immediately:

```
File "/app/modbus_server.py", line 310, in <module>
    register=CONFIG["registers"]["inputRegister"],
KeyError: 'inputRegister'
```

modbus-server 1.4.1 indexes all four register blocks directly, so a config
without `inputRegister` fails at boot — after the registers are read, but before
the socket is opened. Compose therefore reported the container as started while
nothing ever listened on the port, and the EVtec emulator kept logging
`lost connection to evtec-mock:5020`.

`quasar_charging_33pct.json` already carried an empty `inputRegister` block,
which is exactly why that mock was unaffected.

## Changes

- **`charger-mocks/configs/evtec_bidipro_33pct.json`** — add the empty
  `inputRegister` block, so both mock configs describe all four register types.
- **`test_charger_emulator_evtec.py`** — test that fails if any of the four
  blocks goes missing again.


ArdJonker (commented):
Self approved

PR #494, opened 2026-09-13
Test the default FlexMeasures server, not the configured one

ArdJonker:
Switching "use other than default server" back off had no effect: the connection test kept hitting the custom URL, failed on credentials that only exist on the other server, and left the asset list empty -- so the dialog refused to save and the user could not get back to the default server at all.

__test_schedule_connection fell back to c.FM_BASE_URL, which is not the default but a mutable global that __initialise_fm_client_settings overwrites at start-up with the *configured* URL. While the user is switching back, that global still holds the custom URL. Fall back to the setting's factory default instead.

Also add the missing f-prefix to four log statements, three of them in test_fm_connection itself, which logged the literal "host: '{host}'" and hid exactly the information needed to spot this.


ArdJonker (commented):
Self approved

PR #496, opened 2026-09-15
359 fase3 evtec

ArdJonker:
## What

Adds support for a second bidirectional charger, the EVtec BiDiPro 10, and
lets the user pick which charger they have. Until now V2G Liberty built a
Wallbox Quasar 1 driver unconditionally.

Existing installations need to do nothing: a settings migration assumes
`wallbox-quasar-1` when the charger is configured but no type is stored,
which was the only possibility before.

## Changes

- **`chargers/evtec_bidipro.py`** — a driver on the hardware-tested ECP4 /
  Modbus 2.0 register map. Follows the structure of `wallbox_quasar_1.py`
  rather than the shape this driver had on `359-add-new-modbus-charger`: the
  same exception and grace-timer machine, the same poll strategy, the same
  eight events with the same arguments and ordering.
- **`chargers/factory.py`** — maps the stored type onto a driver class.
  `v2g_app` builds from the stored type at start-up; changing the type in the
  settings swaps the running driver in place, without a restart.
- **Charger-type selector** in the settings dialog, with a default port per
  type, and the type shown on the charger card.
- **A charger can occupy more than one phase.** `connected_to_phase` is now a
  set of phases, so a 3-phase EVtec fits, and a 2-phase charger on `[2, 3]`
  would too. A refused save is no longer silent.
- **`ChargerPhaseDetector` refuses a second concurrent run.** It pauses the
  charge mode and commands power; two runs would leave the user with a
  charger doing nothing, and a phase result that cannot be trusted.

Also included, found while testing this branch: the escalation reason no
longer arrives as AppDaemon's kwargs dict, and the charge-mode reset log line
names the mode actually left.

## Not in this PR

How V2G Liberty *communicates* charger faults — a refused discharge is
currently silent, and the error card is Quasar-specific. That is a review of
shared behaviour affecting both chargers, so it sits in a stacked PR on top
of this one: `359-fase3-evtec-error-ux`. Merge this one first.


ArdJonker (commented):
Self approved

PR #497, opened 2026-09-15
359 fase3 evtec error ux

ArdJonker:
## What

A charger can refuse to do its job in three ways, and V2G Liberty said
nothing useful about any of them.

**A refused discharge was completely silent.** The car is connected, charging
works, and the schedule simply never earns anything back. Pressing "Max
discharge now" appeared to do nothing at all.

**An un-recoverable error blamed the wrong thing.** Modbus being unreachable
and the charger reporting a fault need opposite first steps from the user,
but both produced "The charger is not responding" through an entity named
`input_boolean.charger_modbus_communication_fault`. Someone whose charger
reports fault code 12 was sent to check their network, by a critical
notification, at night.

**The recovery steps were cut off** halfway through the numbered list, and
they told every user to open the Wallbox app.

## Changes

- `sensor.charger_problem` (`none` / `communication` / `charger_error`)
  replaces the boolean; the driver's `reason` now reaches `main_app`.
- The error card moves to its own full-width grid at the top, stops clipping,
  and its first step follows both the problem and the charger type.
- New `discharge_refused` event carrying a reason code. `main_app` turns it
  into `sensor.discharge_refused`, an alert under the chart next to "no new
  schedule", and a notification — immediate for a manual request, after an
  hour for a scheduled one.
- The refusal clears itself as soon as its cause is gone, and discharging is
  retried; neither the schedule nor the max-discharge prognosis is drawn while
  a refusal stands.

## Note

The remedy texts are provisional pending confirmation from the authors of the
EVtec Modbus contract. They live in one table per concern, so replacing them
is a data change.


ArdJonker (commented):
Self approved

PR #498, opened 2026-09-16
359 fase3b Charger settings: one save for the whole flow, and the phase on request

ArdJonker:
Rounds off the charger selection flow after #496 and #497, with the findings
from testing those.

**The charger settings are saved once, at the end** (`5e0c8ce`). They used to be
stored — and the driver swapped — before the phase step was shown, so backing
out of that step left a new charger type with the phase of the charger it
replaced: a one-phase Quasar carrying the EVtec's `[1, 2, 3]`. Silent and
lasting, because `data_monitor` then spreads the charge power over three phases
while the charger sits on one. The dialog now collects everything and writes it
in one call at the end; Back or the cross leaves nothing changed. The phase is
validated before a single setting is stored, and **the separate
`save_charger_phase` event is gone** — two writers of the same setting, each
with its own validation, is how the two could drift apart. Because detection
measures the charger the running driver talks to, the phase step explains
itself instead of offering the button while changes are unsaved.

**The phase is asked for, not detected on connect** (`9cba365`). Detection
forces the charge mode to Stop and charges the car to measure; doing that
unasked on every connect is a behaviour choice, and the trigger lost the race
with the grid connection settings on a restart anyway. The manual path already
exists behind a button in the phase step, so the automatic one is dropped and
the card's warning now says where to set it.

**Three smaller fixes.** The unrecoverable-error card is full width —
`column_span: 3` on the section, `layout_options` on the conditional (`7c774a7`).
The charger status card no longer scrolls: content starting with `<ha-icon>`
left a stray empty paragraph after the table (`77c6d31`). And the stored phase
now lights up in the selection: the setting holds a phase set (`[2]`) while the
radio buttons compare against a bare number (`29019d7`, predates this branch).


ArdJonker (commented):
Self approved

PR #499, opened 2026-09-18
Recover by itself once an unusable charger is back

ArdJonker:
Giving up on a charger stopped the polling and left
the charge mode on Stop, with no way back: the driver no longer read the
charger, so it could not notice the fault was over. Recovery meant restarting
both charger and app — which is what the error card prescribed.

A shared `RecoveryProbe` now re-checks the charger every minute after an
un-recoverable error and stands down at the first healthy answer. It sits next
to the drivers rather than in the base classes (those are deliberately thin),
so the timer administration — single-flight, the slot claimed before the first
await, released when arming fails — exists once instead of twice.

**The health check covers both ways a charger becomes unusable:** one raw read
of the state and error registers, past the exception state machine so it arms
no grace timer. Healthy means reachable *and* not in an error state *and* no
error bits — for the charger that stays reachable while reporting a fault, a
live connection proves nothing.

**The charge mode is restored only if this app forced Stop.** `main_app`
remembers what it overrode. Always back to Automatic, never to the boost that
may have been on (deliberate, momentary, and hours may have passed); an unknown
mode counts as Stop, since overriding a deliberate Stop is worse than asking
for a click.

**The error card** dropped the steps that are now pointless (restart V2G
Liberty, wait, switch back to Automatic) and says the app picks up by itself.
It gained a check of power and network — the first thing to look at when a
charger does not respond.


ArdJonker (commented):
Self approved

PR #500, opened 2026-09-21
Keep starting when Home Assistant does not know an entity yet

ArdJonker:
The add-on copies its package to Home Assistant on every start but never
restarts it, so the first boot after a release that adds a helper runs without
that entity. `hass.get_state()` returns None, and `__process_setting` read from
it unconditionally — a `TypeError` that aborts `kick_off_settings` and with it
the whole app: no charger, no schedules, no charging, until someone happens to
restart Home Assistant.

**This release hits it.** 0.8.3 adds `input_text.charger_type` (Fase 3, #496) —
the only new entity in it, checked against `main` — and every existing user
reads that setting at start-up.

An unknown entity is now treated as an empty one: the stored value is still
returned and only what is read *from* the entity is skipped. Each missing
entity is reported once, naming the restart that resolves it. Four tests cover
it: the stored value comes back, the warning appears exactly once, nothing
stored and no factory default returns `""` rather than raising, and the normal
path with a known entity is untouched.

The substitute is an empty entity shape rather than `{}`, so the method still
answers `""` — its documented "nothing" — instead of leaking a `None` into the
constants.


ArdJonker (commented):
Self approved

PR #501, opened 2026-09-21
Stop a Home Assistant restart from raising false alarms

ArdJonker:
Restarting Home Assistant produced two warnings that told the user to fix
something that was not wrong. Both are cards-only; base `dev`.

### "App not configured correctly", with an empty list

Not the settings flags — those survive a restart. It was the two connection
sensors: V2G Liberty writes them with `set_state`, so Home Assistant does not
restore them. After a restart they read `'unknown'`, then `''`, until the app
rewrites them. `'unknown'` counted as a connection error, so the dialog opened;
three seconds later the state was `''` (falsy), the list was empty, and the
dialog stayed put. Measured: `'unknown'` at 07:37:45, `''` at 07:37:48,
`'Successfully connected'` at 07:37:50.

- States that carry no information (`''`, `'unknown'`, `'unavailable'`) no
  longer count as a connection error.
- Deciding whether to warn and rendering what is wrong now share one
  implementation, instead of two copies that could disagree.
- The dialog closes itself once there is nothing left to report, so it can never
  stand there empty.
- A problem has to persist two seconds before the dialog opens at all.

### "Connection with V2G Liberty lost, please restart the add-on"

A Home Assistant restart takes the app with it: AppDaemon waits five seconds
before reconnecting and only then re-initialises the apps, so for a few seconds
Home Assistant is up while nothing answers the ping (HA back at 09:52:07, the
app serving again at 09:52:11).

- The card waits twenty seconds before it alarms — a little over twice the
  measured window, and short enough to report a real outage promptly.
- It stays silent altogether while the frontend has no connection to Home
  Assistant, asking the socket rather than the `hass` object: that object keeps
  claiming it is connected, because the frontend stops updating cards while it
  is away.

Both keep recovering instantly — the checks repeat every second, so the moment
the app answers again the toast disappears.


ArdJonker (commented):
Self approved

